PR #5, opened 2026-02-08
fix: filesystem incorrectly read/writes, rewrite POSIX memory backend to enable rendering

Loreaxe:
Fixed Linux read write issues with files. 

- Make PosixFileHandle::Read/Write handle errors cleanly:
  * return false on pread/pwrite failure
  * avoid writing garbage into out_* when the syscall fails
- Populate FileInfo::total_size and type consistently in GetInfo

=====================================

Possible fix for binutils by adding POSIX +x flags to files.

=====================================

Rework the POSIX memory backend to more closely match Windows-style
reserve/commit behavior and avoid unsafe MAP_FIXED usage.

Key changes:
- Implement reserve vs commit semantics using PROT_NONE + mprotect
- Prefer MAP_FIXED_NOREPLACE and refuse unsafe fixed-address mappings
- Emulate commit-on-reserve when mappings already exist
- Distinguish decommit vs release (mprotect+MADV_DONTNEED vs munmap)
- Add Linux /proc/self/maps introspection for QueryProtect and span inference
- Tighten file mapping behavior (page-aligned offsets, fixed-base validation)

image


Roxxsen (dismissed):
Builds and also resolves issues with binutils executable flag.

PR #9, opened 2026-02-09
fix: Add POSIX execute to binutils files.

Loreaxe:
Adding POSIX execute to binutils files should solve users issues during compile.


Roxxsen (approved):
Thank you, simple and clean.

PR #13, opened 2026-02-10
Fix Assorted POSIX Issues

tomcl7:
- Incomplete struct fill in GetInfo, namely the total size.
- Rewrite the host memory management for POSIX to properly emulate Windows VirtualAlloc semantics.


Loreaxe (approved):
Working like a charm.

PR #15, opened 2026-02-11
feat: add vsubuws instruction

Natan822:
Adds codegen support for the vsubuws instruction:
- vsubuws — vector subtract unsigned word saturate

Files changed:
- src/codegen/builders.h — added declaration for vsubuws
- src/codegen/builders/vector.cpp — added implementation for vsubuws
- src/codegen/instruction_dispatch.cpp — added dispatch entry mappings for vsubuws


tomcl7 (approved):
Looks good to me. 

- Algorithm is `A - min(A, B) yields max(A - B, 0)` for unsigned, which is exactly what vsubuws should do. 
- Operand mapping `([0]=dest, [1]=A, [2]=B)` checks out with the four format args being `[0], [1], [1], [2]`.

PR #42, opened 2026-02-16
Use context from thread_state on thread execute

### LINKED ISSUE #41: Threads not using context from thread_state_
In `XThread::Execute`, a PPCContext is created locally and passed to the guest code. This way, the per-thread context accessible from `thread_state_->context()` which is relevant to some kernel functions such as KeSetCurrentStackPointer does not reflect the true context at the time of being called.

Mystixor:
Threads should be started with a context taking from thread_state_, so that later on the context can be accessed from kernel functions. Fixes #41.

### DIFF
--- a/src/kernel/xthread.cpp
+++ b/src/kernel/xthread.cpp
@@ -536,7 +536,7 @@ void XThread::Execute() {
   }
 
   // Set up the PPCContext for execution
-  PPCContext ctx{};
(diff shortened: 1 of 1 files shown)


tomcl7 (changes requested):
See my remarks in https://github.com/rexglue/rexglue-sdk/issues/50

Issue #30, opened 2026-02-14
Missing opcode: cror


Mystixor:
`cror 4*cr1+eq,lt,4*cr6+eq`

4c c0 d3 82

Issue #31, opened 2026-02-14
Missing opcode: crorc


Mystixor:
`crorc eq,4*cr1+eq,4*cr5+lt`

4C 46 A3 42

PR #51, opened 2026-02-17
Opcode cr

CradleKing24:
cror - Condition Register OR (x’4C00 0382’)
`CR[crbD] ← CR[crbA] | CR[crbB]`

crorc - Condition Register OR with Complement (x’4C00 0342’)
`CR[crbD] ← CR[crbA] | ¬ CR[crbB]`

Fixes https://github.com/rexglue/rexglue-sdk/issues/30 and https://github.com/rexglue/rexglue-sdk/issues/31


tomcl7 (changes requested):
The change to emitCRBitOperation needs to be a singular commit. The implementation of cror also needs to be its own commit. 

I will accept it if you squash the seemingly redundant 2 commits for the helper and cror.

PR #61, opened 2026-02-19
Implement vsubuwm

helgot:
Implement vsubuwm instruction.

Issue: https://github.com/rexglue/rexglue-sdk/issues/60

### DIFF
--- a/src/codegen/builders/vector.cpp
+++ b/src/codegen/builders/vector.cpp
@@ -323,6 +323,12 @@ bool build_vsubuhm(BuilderContext& ctx)
     return true;
 }
 
+bool build_vsubuwm(BuilderContext& ctx)
+{
+    ctx.emit_vec_int_binary("sub_epi32", "u32");
+    return true;
+}
+
 bool build_vmaxsw(BuilderContext& ctx)
 {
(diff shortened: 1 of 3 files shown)


src/codegen/builders/vector.cpp
```diff
@@ -323,6 +323,12 @@ bool build_vsubuhm(BuilderContext& ctx)
     return true;
 }
 
+bool build_vsubuwm(BuilderContext& ctx)
+{
+    ctx.emit_vec_int_binary("sub_epi32", "s32");
```
tomcl7: Type accessor is (technically) wrong.

In practice, for modular subtraction the result is the same regardless of signed/unsigned interpretation (two's complement), so this likely doesn't produce wrong results. But for consistency and correctness of intent, it should probably be `u32` to match the name and convention (`unsigned word modulo`).

tomcl7 (changes requested):
Type accessor looks wrong.

src/codegen/builders/vector.cpp
helgot: Ah, yes, thank you for the explanation. I have updated the type accessor to `u32`.

PR #63, opened 2026-02-19
Implement vsubuhs

helgot:
Implement vsubuhs instruction

Issue: https://github.com/rexglue/rexglue-sdk/issues/62

### DIFF
--- a/src/codegen/builders/vector.cpp
+++ b/src/codegen/builders/vector.cpp
@@ -317,6 +317,12 @@ bool build_vsubuws(BuilderContext& ctx)
     return true;
 }
 
+bool build_vsubuhs(BuilderContext& ctx)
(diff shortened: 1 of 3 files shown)


src/codegen/builders/vector.cpp
```diff
@@ -317,6 +317,12 @@ bool build_vsubuws(BuilderContext& ctx)
     return true;
 }
 
+bool build_vsubuhs(BuilderContext& ctx)
+{
+    ctx.emit_vec_int_binary("subs_epu16", "u8");
```
tomcl7: Type accessor is wrong. It should be `u16` for a half sat.

tomcl7 (approved):
Looks good, thank you!

PR #71, opened 2026-02-20
Replace date:: library with C++23 std::chrono::

tomcl7:
## Summary
- Mechanical `date::` --> `std::chrono::` migration across 7 source files
- Add 17 chrono unit tests pinning epoch, FILETIME round-trip, calendar decomposition/recomposition, and weekday behavior
- Remove `thirdparty/date` submodule and build dependency


Loreaxe (approved):

Roxxsen (approved):
Compiles, generates and runs iruka

PR #72, opened 2026-02-20
add NtCancelIoFile

SamuraiOndo:
this is needed for Tornado Outbreak, and probably many other games


src/kernel/xboxkrnl/xboxkrnl_io.cpp
```diff
@@ -806,3 +815,4 @@ GUEST_FUNCTION_HOOK(__imp__IoDismountVolumeByName, rex::kernel::xboxkrnl::IoDism
 GUEST_FUNCTION_HOOK(__imp__IoSynchronousDeviceIoControlRequest, rex::kernel::xboxkrnl::IoSynchronousDeviceIoControlRequest_entry)
 GUEST_FUNCTION_HOOK(__imp__StfsCreateDevice, rex::kernel::xboxkrnl::StfsCreateDevice_entry)
 GUEST_FUNCTION_HOOK(__imp__StfsControlDevice, rex::kernel::xboxkrnl::StfsControlDevice_entry)
+GUEST_FUNCTION_HOOK(__imp__NtCancelIoFile, rex::kernel::xboxkrnl::NtCancelIoFile_entry)
```
tomcl7: Fix line ending. Should be LF through the whole document with an EOF LF.

Otherwise, looks good.

SamuraiOndo: my apologies, will do if i contribute anymore

PR #94, opened 2026-02-22
SDK Overhaul - rexsystem, PPC types, Platform split

tomcl7:
## Summary

Consolidates `rex::kernel` and `rex::runtime` into a unified `rex::system` library, splits monolithic platform code into `_win`/`_posix` pairs, rewrites the PPC guest type system, adds dependency injection for subsystems, introduces `ReXApp` as a base class with built-in ImGui overlays, and adds submodule consumption support.

---

### Library restructure

- New `rexsystem` static library absorbs all 24 kernel sources and the entire `rexruntime` library
- `rexkernel` reduced to only XAM/xboxkrnl export implementations
- `rexruntime` dissolved entirely, `src/runtime/CMakeLists.txt` deleted
- Added `LogCategory::System` with `REXSYS_*` logging macros

### Platform split

- `threading.cpp`, `string.cpp`, `clock.cpp` each split into `_win.cpp`/`_posix.cpp`
- New platform-split files: atomics, dynlib, math intrinsics, SEH, sockets
- Platform headers (`win.h`, `linux.h`) moved from public `include/` to private `src/core/`
- ~200 lines of inline intrinsics moved out of headers into `.cpp` files
- New cross-platform abstractions: `dynlib.h`, `socket.h`, `platform::lib_names`

### PPC refactoring

- New `include/rex/ppc/` header library replaces `include/rex/runtime/guest/`
  - `ppc/types.h` - `PPCPointer<T>`, `PPCValue<T>`, register types, type aliases
  - `ppc/context.h` - `PPCContext`, `PPCFuncMapping`, register accessors, `ppc_setjmp`/`ppc_longjmp`
  - `ppc/function.h` - `PPC_HOOK`, `HostToGuestFunction`/`GuestToHostFunction`, argument translation, type traits
  - `ppc/exceptions.h` - cross-platform SEH (`initialize_seh()`, `SEH_TRY`/`SEH_CATCH` macros)
  - `ppc/memory.h` - guest memory access utilities
  - `ppc/detail/fpscr.h` - FPSCR register definitions
  - `ppc/detail/seh.h` - internal SEH state
- `GUEST_FUNCTION_HOOK` renamed to `PPC_HOOK`
- Type aliases renamed from Win32-style to `ppc_` prefix (see breaking changes - PPC Type Aliases)
- `HostToGuestFunction` now sets `g_current_ppc_context` and `g_memory_base` thread-locals, enabling `GuestToHostFunction` callbacks from kernel exports

**`PPCValue<T>`** - new wrapper for scalar guest values:

- Replaces raw typedefs like `typedef uint32_t dword_t`
- Wraps a value with `.value()` accessor and implicit conversion to `T`
- Full arithmetic, bitwise, and comparison operator set
- Templated comparison/arithmetic against any `std::convertible_to<T>` type
- `explicit operator bool()` to prevent accidental integer-to-bool conversion
- `fmt::formatter` specialization for direct use in log macros

**`PPCPointer<T>`** - rewritten guest pointer type:

- Tracks both host pointer (`T*`) and guest address (`uint32_t`) simultaneously
- `operator T*()` - implicit conversion to host pointer (most common use)
- `operator uint32_t()` - now `explicit` (use `.guest_address()` instead)
- `explicit operator bool()` - null check
- `.value()` - dereferences and byte-swaps if `T` is a `be<>` type
- `.as<U>()` - reinterpret cast to another type
- `.as_array<U>()` - cast to `be<U>*` for array access
- `.Zero()` / `.Zero(size)` - zero-fill the pointed-to memory
- Pointer arithmetic (`+`, `-`) adjusts both host pointer and guest address
- Specializations for `void`, `char` (string_view `.value()`), and `char16_t` (u16string_view `.value()`)

### Namespace & include migration

(see breaking changes - Include Paths)

### ReXApp base class

`rex::ReXApp` is a new base class that absorbs all boilerplate for recompiled windowed applications: runtime setup, window creation, ImGui wiring, module launch, and shutdown.

Consumer projects inherit `ReXApp` and optionally override virtual hooks:

- `OnPreSetup(RuntimeConfig&)` - modify backend config before `Runtime::Setup()`
- `OnPostSetup()` - called after runtime init, before window creation
- `OnCreateDialogs(ImGuiDrawer*)` - add custom ImGui dialogs
- `OnShutdown()` - release custom resources before cleanup

The constructor takes a `PPCImageInfo` struct containing PPC image layout from the generated config header:

```cpp
struct PPCImageInfo {
  uint32_t code_base;
  uint32_t code_size;
  uint32_t image_base;
  uint32_t image_size;
  const PPCFuncMapping* func_mappings;
};
```

Each consumer project generates a static `Create()` factory method on its subclass (via `rexglue init` / `rexglue migrate`):

```cpp
class IrukaApp : public rex::ReXApp {
 public:
  using rex::ReXApp::ReXApp;

  static std::unique_ptr<rex::ui::WindowedApp> Create(
      rex::ui::WindowedAppContext& ctx) {
    return std::unique_ptr<IrukaApp>(new IrukaApp(ctx, "iruka",
        {PPC_CODE_BASE, PPC_CODE_SIZE, PPC_IMAGE_BASE,
         PPC_IMAGE_SIZE, PPCFuncMappings}));
  }
};
REX_DEFINE_APP(iruka, IrukaApp::Create)
```

### Built-in overlays

ReXApp includes three ImGui overlay dialogs by default, wired up automatically:

| Overlay | Header | Default key | Description |
|---------|--------|-------------|-------------|
| Debug | `<rex/ui/overlay/debug_overlay.h>` | F3 | Host/guest FPS and frame timing |
| Console | `<rex/ui/overlay/console_overlay.h>` | Backtick | Log viewer with level/category filters and cvar command input |
| Settings | `<rex/ui/overlay/settings_overlay.h>` | F4 | Live cvar editor with search, categorized view, and save-to-config |

All three are constructed in `ReXApp::OnInitialize()` after ImGui setup and destroyed in `OnDestroy()`. No consumer code required.

The debug overlay accepts an optional `FrameStatsProvider` callback for guest-side frame stats. Without it, only host (ImGui) FPS is shown.

The console overlay uses `rex::LogCaptureSink` (`<rex/log_capture.h>`), a header-only thread-safe ring-buffer spdlog sink (2048 entries) that ReXApp attaches to all loggers on startup. Categories are discovered dynamically from log entries.

### Keybind system

New global keybind registry in `<rex/ui/keybinds.h>`:

- `RegisterBind(name, default_key, description, callback)` - registers a keybind backed by a cvar in the "Keybinds" category
- `UnregisterBind(name)` - removes a keybind
- `ProcessKeyEvent(KeyEvent&)` - dispatches key events to matching binds

Each overlay self-registers its toggle keybind in its constructor and unregisters in its destructor. ReXApp dispatches key events via `WindowInputListener::OnKeyDown`.

Keybinds are editable at runtime through the Settings overlay (they appear as cvars in the "Keybinds" category).

### New APIs

- `IAudioSystem`, `IGraphicsSystem`, `IInputSystem` - abstract interfaces for dependency injection
- `RuntimeConfig` struct with factory-based subsystem setup
- `Runtime::Setup(bool tool_mode)` → `Runtime::Setup(RuntimeConfig config = {})`
- `InputSystem` decoupled from `KernelState`, ownership moved to `Runtime`

### SDK consumption

The SDK now supports two consumption modes:

**Submodule** (new) - set `REXSDK_DIR` to the rexglue-sdk source tree:
```cmake
set(REXSDK_DIR "thirdparty/rexglue-sdk")
```

**Installed** - `find_package(rexglue)` via install prefix, env var, or CMake user package registry (existing behavior).

The generated consumer CMakeLists handles both:
```cmake
set(REXSDK_DIR "" CACHE PATH "Path to rexglue-sdk source tree")
if(REXSDK_DIR)
    add_subdirectory("${REXSDK_DIR}" rexglue-sdk)
else()
    find_package(rexglue QUIET CONFIG)
    if(NOT rexglue_FOUND)
        message(FATAL_ERROR "ReXGlue SDK not found. Either:\n"
            "  - Set REXSDK_DIR to the rexglue-sdk source tree\n"
            "  - Install the SDK package and ensure it is on CMAKE_PREFIX_PATH")
    endif()
endif()
```

All internal `CMAKE_SOURCE_DIR` references replaced with `REXGLUE_ROOT` (set to `CMAKE_CURRENT_SOURCE_DIR` in the SDK's top-level CMakeLists) so paths resolve correctly in both modes.

### `rexglue migrate` command

New command regenerates SDK-managed files (`CMakeLists.txt`, `src/main.cpp`) to match the current SDK template. Brings existing projects up to date with ReXApp pattern, PPCImageInfo, and new CMake discovery logic.

### Tooling

- `.clang-format` - Google-based, 2-space indent, 100 columns, K&R braces
- `.clang-tidy` - bugprone, performance, naming convention enforcement
- `.cmake-format.yaml` - CMake formatting rules
- `.editorconfig` / `.gitattributes` - LF line endings, UTF-8
- `scripts/format.{sh,ps1}` / `scripts/lint.{sh,ps1}` - cross-platform helpers
- `.github/workflows/lint.yaml` - CI blocks PRs on format or lint violations
- `CONTRIBUTING.md` - links to wiki contributing guide

---

## Bug fixes

- `install_signal_handler` set the installed flag on failure instead of success
- `affinity_mask()` passed a pointer as the mask value to `SetThreadAffinityMask`
- `1 << i` in POSIX affinity loop caused UB for `i >= 31` (now `1ULL << i`)
- Unbalanced parenthesis in `PPC_LOOKUP_FUNC` macro default
- `LOG_RUNTIME_DEBUG` undefined in `PPC_STUB_LOG`/`PPC_STUB_RETURN` (now `REXKRNL_DEBUG`)
- `HostToGuestFunction` never set `g_current_ppc_context`/`g_memory_base`, causing silent failure for `GuestToHostFunction` callbacks
- `g_seh_initialized` never set on Windows
- `!REX_PLATFORM_WIN32` static assert warned with `-Wundef` on non-Windows
- Non-atomic read race in `__sync_val_compare_and_swap` (now `__atomic_exchange_n`)
- `RuntimeConfig` move semantics broken due to forward-declared interface types
- Incorrect `emit_vec_int_binary()` in several codegen vector instruction builders

---

## Breaking changes

The primary motivation for this overhaul is breaking a circular dependency between `rexkernel` and `rexgraphics`. The old `rexkernel` conflated two responsibilities, system emulation (KernelState, kernel objects, memory management) and kernel exports (xboxkrnl_video, xam_input, etc.). The exports need graphics/audio/input, but the system layer doesn't. Separating them into `rexsystem` (system emulation) and `rexkernel` (exports only) eliminates the cycle and produces a clean dependency graph where all arrows point downward.

The naming changes were a secondary goal. The original codebase inherited conventions from Xenia that don't align with how a statically recompiled project uses the SDK. When dealing with a fragile host/guest boundary, clarity matters. Names like `ppc_u32_t`, `rex::system`, and `include/rex/ppc/` make the origin and purpose obvious at a glance.

### Include paths

Remaining moves group related functionality (`chrono/`, `crypto/`) and shorten verbose names (`debugging.h` → `dbg.h`).

| Before | After |
|--------|-------|
| `<rex/kernel/kernel_state.h>` | `<rex/system/kernel_state.h>` |
| `<rex/kernel/xthread.h>` | `<rex/system/xthread.h>` |
| `<rex/kernel/xmemory.h>` | `<rex/system/xmemory.h>` |
| `<rex/runtime/processor.h>` | `<rex/system/processor.h>` |
| `<rex/runtime/xex_module.h>` | `<rex/system/xex_module.h>` |
| `<rex/runtime/guest/*.h>` | `<rex/ppc/*.h>` |
| `<rex/time/chrono.h>` | `<rex/chrono/chrono.h>` |
| `<rex/sha256.h>` | `<rex/crypto/sha256.h>` |
| `<rex/debugging.h>` | `<rex/dbg.h>` |
| `<rex/byte_order.h>` | `<rex/types.h>` (merged) |
| `<rex/profiling.h>` | `<rex/dbg.h>` (merged) |
| `<rex/xxhash.h>` | deleted |
| `<rex/xbox/xbox.h>` | deleted |

### Namespaces

- `rex::kernel` → `rex::system` (all kernel state, xobject types, utilities)
- `rex::runtime` → `rex::system` (processor, modules, threads)
- `rex::crypto` - new namespace for `sha256`

### PPC type aliases

The old names (`dword_t`, `lpvoid_t`, `lpstring_t`) were inherited from Xenia's Win32 naming conventions. In the context of an SDK consumed by recompiled games, these names are ambiguous - `dword_t` could be any 32-bit value, and `lpvoid_t` looks like a Windows API type. The `ppc_` prefix makes it immediately clear these are PPC guest types that go through calling convention translation, not native host types.

| Before | After | Underlying type |
|--------|-------|-----------------|
| `dword_t` | `ppc_u32_t` | `PPCValue<rex::u32>` |
| `qword_t` | `ppc_u64_t` | `PPCValue<rex::u64>` |
| `int_t` | `ppc_i32_t` | `PPCValue<rex::i32>` |
| `float_t` | `ppc_f32_t` | `PPCValue<rex::f32>` |
| `double_t` | `ppc_f64_t` | `PPCValue<rex::f64>` |
| `function_t` | `ppc_fn_t` | `PPCValue<rex::u32>` |
| `unknown_t` | `ppc_unknown_t` | `PPCValue<rex::u32>` |
| `lpvoid_t` | `ppc_pvoid_t` | `PPCPointer<void>` |
| `lpdword_t` | `ppc_pu32_t` | `PPCPointer<rex::be_u32>` |
| `lpstring_t` | `ppc_pchar_t` | `PPCPointer<char>` |
| `pointer_t<T>` | `ppc_ptr_t<T>` | `PPCPointer<T>` |
| `dword_result_t` | `ppc_u32_result_t` | `rex::u32` |
| `pointer_result_t` | `ppc_ptr_result_t` | `rex::u32` |
| `X_HRESULT_result_t` | `ppc_hresult_result_t` | `rex::i32` |

### PPC macros and functions

- `GUEST_FUNCTION_HOOK(sub, fn)` → `PPC_HOOK(sub, fn)` - shorter, consistent with the `ppc_` naming convention, and `GUEST` was vague (guest of what?)
- `PPCPointer<T>` implicit conversion to `uint32_t` is now `explicit` - the old implicit conversion caused ambiguous overload resolution when both `operator T*()` and `operator uint32_t()` were candidates. Use `.guest_address()` or `static_cast<uint32_t>(ptr)` instead

### Runtime initialization

This is the key change that breaks the circular dependency. Previously, `Runtime::Setup(bool tool_mode)` hardcoded which graphics/audio/input backends to use, which meant `rexkernel` had to link `rexgraphics`, `rexaudio`, and `rexinput` directly - creating the cycle. `RuntimeConfig` inverts the dependency: the caller injects backends via abstract interfaces (`IGraphicsSystem`, `IAudioSystem`, `IInputSystem`), and `rexsystem` doesn't know or care which implementations it gets. Backend selection (`#if REX_HAS_D3D12` / `#if REX_HAS_VULKAN`) moves out of the SDK and into `ReXApp`.

There are two overloads:

- `Setup(RuntimeConfig config = {})` - standalone runtime (Xenia-style)
- `Setup(code_base, code_size, image_base, image_size, func_mappings, RuntimeConfig config = {})` - recompiled game with PPC dispatch table

**What changed under the hood:**

- `InputSystem` no longer created by `KernelState` - owned by `Runtime` via `input_factory`
- Graphics/audio/input initialization is gracefully skipped when the corresponding factory is not provided
- Failed audio or input init logs a warning and continues (non-fatal); failed graphics init returns an error

### CMake discovery

The generated consumer `CMakeLists.txt` no longer auto-detects the SDK via env vars or registry. Two explicit modes:

- `REXSDK_DIR` → `add_subdirectory()` (submodule)
- `find_package(rexglue)` → installed SDK

If neither is configured, CMake emits a `FATAL_ERROR` with instructions.

### DIFF
--- a/src/rexglue/CMakeLists.txt
+++ b/src/rexglue/CMakeLists.txt
@@ -19,4 +19,14 @@ target_link_libraries(rexglue
         rexcodegen
 )
 
+# TODO: fix the circular dependency with rexsystem and rexkernel (init/ordinals)
+if(UNIX)
+    target_link_libraries(rexglue PRIVATE
+        -Wl,--start-group
+        rexsystem
+        rexkernel
+        -Wl,--end-group
+    )
+endif()
+
 add_executable(rex::rexglue ALIAS rexglue)


YoshiCrystal9 (approved):

src/ui/keybinds.cpp
```diff
@@ -0,0 +1,213 @@
...
+/* ---- Bind registry ---- */
+
+struct BindEntry {
+  std::string name;
+  std::string current_key;
+  std::function<void()> callback;
+};
+
+static std::mutex g_binds_mutex;
+static std::vector<BindEntry> g_binds;
```
rapidsamphire: Mentioned this in the discord but I'd look into using std::deque for the bind storage here instead.

PR #179, opened 2026-03-04
Implement vavgsw instruction

helgot:
Implement vavgsw (Vector Average Signed Word) instruction, and rex::simde_mm_avg_epi32 helper.

### DIFF
--- a/include/rex/ppc/memory.h
+++ b/include/rex/ppc/memory.h
@@ -360,6 +360,13 @@ inline simde__m128i simde_mm_avg_epi16(simde__m128i a, simde__m128i b) {
                             simde_mm_avg_epu16(simde_mm_xor_si128(c, a), simde_mm_xor_si128(c, b)));
 }
 
+// Signed 32-bit average
+inline simde__m128i simde_mm_avg_epi32(simde__m128i a, simde__m128i b) {
+  simde__m128i sum = simde_mm_add_epi32(simde_mm_srai_epi32(a, 1), simde_mm_srai_epi32(b, 1));
+  return simde_mm_add_epi32(sum,
+                            simde_mm_and_si128(simde_mm_or_si128(a, b), simde_mm_set1_epi32(1)));
+}
+
 // Convert unsigned 32-bit integers to floats
 inline simde__m128 simde_mm_cvtepu32_ps_(simde__m128i src1) {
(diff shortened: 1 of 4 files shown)


src/codegen/instruction_dispatch.cpp
```diff
@@ -468,6 +468,7 @@ static const std::unordered_map<int, Builder>& GetDispatchTable() {
       //=====================================================================
       {PPC_INST_VAVGSB, build_vavgsb},
       {PPC_INST_VAVGSH, build_vavgsh},
+      {PPC_INST_VAVGSH, build_vavgsw},
```
tomcl7: The dispatch key is wrong. It should be `PPC_INST_VAVGSW`

include/rex/ppc/memory.h
```diff
@@ -360,6 +360,13 @@ inline simde__m128i simde_mm_avg_epi16(simde__m128i a, simde__m128i b) {
                             simde_mm_avg_epu16(simde_mm_xor_si128(c, a), simde_mm_xor_si128(c, b)));
 }
 
+// Signed 32-bit average
+inline simde__m128i simde_mm_avg_epi32(simde__m128i a, simde__m128i b) {
+  simde__m128i sum = simde_mm_add_epi32(a, b);
+  sum = simde_mm_add_epi32(sum, simde_mm_set1_epi32(1));
+  return simde_mm_srai_epi32(sum, 1);
+}
+
```
tomcl7, line 369: There is a bug here. `simde_mm_add_epi32` wraps on overflow for large same-sign inputs.  a + b silently wraps before the shift, producing wrong results.

The AltiVec PPC ISA specifies that each element is added producing a 33-bit signed-integer sum, incremented by 1, with the high-order 32 bits taken as the result. An overflow-safe formula like `(a >> 1) + (b >> 1) + ((a | b) & 1)` would be correct. [You can read more here on page 37.](https://github.com/rexglue/rexglue-sdk/blob/main/docs/ppc/altivec_instructions.pdf)

tomcl7 (changes requested):

include/rex/ppc/memory.h
helgot: Thank you for the pointer. Corrected that now.

src/codegen/instruction_dispatch.cpp
helgot: Corrected.

PR #181, opened 2026-03-05
Use thread_state Context as Execution Context in XThread::Execute

### LINKED ISSUE #41: Threads not using context from thread_state_

tomcl7:
## Summary

XThread::Execute was creating a throwaway PPCContext on the stack instead of using the thread's own thread_state_->context(). This meant the execution context was missing pre-initialized state (r1/stack pointer, r13/PCR, kernel_state) that ThreadState already sets up in its constructor, and any state written during execution was lost when the local went out of scope.

 Fix switches to the thread's existing context, removing the redundant manual initialization of r1, r13, kernel_state, and the local PPCContext declaration.

Fixes #41

### DIFF
--- a/src/system/xthread.cpp
+++ b/src/system/xthread.cpp
@@ -535,47 +535,33 @@ void XThread::Execute() {
     return;
   }
 
-  // Set up the PPCContext for execution
-  PPCContext ctx{};
+  auto* ctx = thread_state_->context();
   uint8_t* base = memory->virtual_membase();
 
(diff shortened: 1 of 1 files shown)


Mystixor (approved):

tomcl7:
Absorbed into #180

PR #182, opened 2026-03-05
Implemented hardware_keycode to Virtualkey translation for Linux.

AllanCat:
While testing with Linux build I found that debug overlay was assigned to wrong keys (F3->Right Arrow, F4->END).
This PR impelments GdkKeyvalToVirtualKey() which should maps keyboard input key correctly.

Also useful for those who are implementing keyboard input support for linux build.


AllanCat:
Closed as already implemented in Fidelity fork [window_gtk.cpp](https://github.com/rexglue/rexglue-sdk/blob/feat/fidelityfx/src/ui/window_gtk.cpp)

Issue #186, opened 2026-03-07
[Bug]: ppu_frsqrte_lut returning invalid values


ihatecompvir:
### Description

Working on the Rock Band 3 recomp I discovered an issue with the `frsqrte` instruction. The LUT was returning `inf` for perfectly normal inputs (e.g. 6.88) which was causing a bunch of obscure bugs. After modifying the generated recompiled code to replace the `ppu_frsqrte_lut` lookups with `1.0 / std::sqrt(x)`, all the remaining graphical and audio issues in Rock Band 3 were resolved. I would PR a fix for this, but I am honestly not 100% sure how to properly fix it or what is really going wrong, so I'll report it here for someone else to fix. It probably causes a lot of issues in other games as well, although it doesn't seem like this instruction is very common.

### Steps to Reproduce

1. Build a game that uses these instructions such as [Rock Band 3](https://github.com/ihatecompvir/band3_recomp)
2. Observe that the hair for the characters in-game is completely invisible, and the startup intro movie has no sound

### Expected Behavior

The LUT should be returning valid floating point values when being passed valid inputs.

### Actual Behavior

The LUT returns invalid floating point values such as `inf` when being passed valid inputs.

### Platform

Windows

tomcl7:
Thank you for this. Looking into it.

PR #194, opened 2026-03-10
Fix Instruction Builder Bugs and Add New Instructions

tomcl7:
### Bug fixes

- **rlwimi**: use 64-bit `x||x` rotate pattern matching rlwinm, fixing wrapped masks that set bits in the high 32-bit half ([core_instructions.pdf](docs/ppc/core_instructions.pdf) 3.3.13 - ROTL32 concatenates the word with itself before rotating)
- **divw/divwu**: zero-extend results to 64 bits via `.u64 = uint32_t(...)` instead of writing `.s32`/`.u32` which left upper bits stale ([core_instructions.pdf](docs/ppc/core_instructions.pdf) 3.3.8 - result placed in low 32 bits, high 32 bits cleared)
- **fctid/fctidz/fctiwz**: add explicit `std::isnan()` checks returning PPC-correct sentinel values (`0x8000000000000000` for 64-bit, `0x80000000` for 32-bit) instead of relying on platform-specific SIMD behavior ([core_instructions.pdf](docs/ppc/core_instructions.pdf) 4.6.7)
- **fres**: add explicit `double()` cast for consistency with other single-precision estimate instructions
- **frsqrte**: replace LUT-based implementation with direct `1.0 / sqrt(x)` computation (#186)
- **blrl**: implement as call-through-LR (save LR, set return address, indirect call) instead of `__builtin_debugtrap()` ([core_instructions.pdf](docs/ppc/core_instructions.pdf) 2.4.1)
- **mtfsf**: apply FM field mask to only update selected FPSCR fields instead of unconditionally overwriting the entire register ([core_instructions.pdf](docs/ppc/core_instructions.pdf) 4.6.5)
- **vspltisb/vspltisw**: sign-extend 5-bit immediate before use, matching the existing vspltish pattern ([altivec_instructions.pdf](docs/ppc/altivec_instructions.pdf) - SIM field is sign-extended to element width)
- **vrlw**: guard right-shift with `sh ? (...) : 0` to prevent UB when shift amount is zero, matching vrlh
- **lvebx/lvehx/lvewx**: implement proper element-sized loads into zeroed vectors instead of aliasing to full 16-byte lvx ([altivec_instructions.pdf](docs/ppc/altivec_instructions.pdf) - loads only the addressed element, other elements undefined)

### New instructions

- **mftbu**: upper 32-bit timebase read (separate PPC_INST from mftb)
- **bdnzlr**: decrement CTR and return if non-zero
- **rldic**: rotate left doubleword immediate then clear
- **rldcl/rldcr**: rotate left doubleword then clear left/right (register rotation amount)
- **lhaux**: halfword algebraic load with update indexed
- **lwaux**: word algebraic load with update indexed
- **stfdux**: store float double with update indexed
- **vcmpgtsb**: signed byte vector compare greater-than (with record form)
- **mfctr**: move from count register
- **mfxer**: move from XER (reconstructs SO/OV/CA bit layout)
- **mtcrf/mtocrf**: move to CR fields with FXM mask (shared implementation handles both)

### DIFF
--- a/src/codegen/builders.h
+++ b/src/codegen/builders.h
@@ -357,6 +357,9 @@ bool build_stfsux(BuilderContext& ctx);
 bool build_stfsx(BuilderContext& ctx);
 
 // Vector loads
+bool build_lvebx(BuilderContext& ctx);
+bool build_lvehx(BuilderContext& ctx);
+bool build_lvewx(BuilderContext& ctx);
 bool build_lvx(BuilderContext& ctx);
 bool build_lvlx(BuilderContext& ctx);
 bool build_lvrx(BuilderContext& ctx);
--- a/src/codegen/builders/memory.cpp
+++ b/src/codegen/builders/memory.cpp
@@ -541,6 +541,32 @@ bool build_lvx(BuilderContext& ctx) {
   return true;
 }
 
+bool build_lvebx(BuilderContext& ctx) {
+  emitVectorEA(ctx);
+  ctx.println("\tmemset({}.u8, 0, 16);", ctx.v(ctx.insn.operands[0]));
+  ctx.println("\t{}.u8[15 - ({} & 0xF)] = *(uint8_t*)PPC_RAW_ADDR({});",
+              ctx.v(ctx.insn.operands[0]), ctx.ea(), ctx.ea());
+  return true;
+}
+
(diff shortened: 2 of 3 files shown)


tomcl7:
I found some issues with this. Going to put into draft while I fix them.

PR #218, opened 2026-03-13
Port over XInput handler from xenia

### LINKED ISSUE #217: [Feature]: Implement XInput controller handling
### Summary

Implement XInput controller handling to allow native instrument and other accessory use

### Proposed Solution

can mostly just be copied from xenia, and then set as a cvar.

### Requirements

1. gamepad types must properly be passed thru to XAM's Xinput functions

### Out of Scope

_No response_

### Modified Dependencies

_No response_

### Acceptance Criteria

- [ ] The DJ Hero turntable is detected by the game DJ Hero 2 and works properly

### Related

PR #204 already has this 100% implemented and tested, I have tested it myself and it works fine.

sanjay900:
Resolves #217 

This lets us use xbox 360 instruments and other peripherals natively

I put selecting the input backend behind a CVAR for now


src/input/sdl/sdl_input_driver.cpp
```diff
@@ -20,8 +20,6 @@
 #include <rex/logging.h>
 #include <rex/ui/virtual_key.h>
 
-REXCVAR_DEFINE_BOOL(guide_button, false, "Input", "Enable guide button pass-through");
```
tomcl7, line 23: Am I missing something or was this an accidental removal?

sanjay900: I moved it to `src/input/input_system.cpp` since that CVAR is used by both handlers

PR #231, opened 2026-03-15
Update PPC_STUB macros so they use PPC_WEAK_FUNC instead of PPC_FUNC

### LINKED ISSUE #230: [Feature]: Make PPC_STUBs weak so they can be overridden
### Summary

Right now, PPC_STUBed functions can't be overridden, which means they can't be implemented in projects using rexglue

### Proposed Solution

Update the macro so that it uses `PPC_WEAK_FUNC` instead of `PPC_FUNC`

### Requirements

- [ ] Change the PPC_STUB macros from using `PPC_FUNC` to `PPC_WEAK_FUNC`

### Out of Scope

_No response_

### Modified Dependencies

_No response_

### Acceptance Criteria

- [ ] Build a project using rexglue-sdk that requires overridding a function that is currently stubbed

### Related

_No response_

sanjay900:
resolves #230

### DIFF
--- a/include/rex/ppc/function.h
+++ b/include/rex/ppc/function.h
@@ -591,21 +591,21 @@ T GuestToHostFunction(const TFunction& func, TArgs&&... argv) {
 
 // Create a simple stub that does nothing
 #define PPC_STUB(subroutine)              \
-  extern "C" PPC_FUNC(subroutine) {       \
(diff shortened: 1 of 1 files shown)


tomcl7 (approved):
Looks good to me.  Thank you!

PR #238, opened 2026-03-16
fix: Fix CI Pipelines after FidelityFX was added

### LINKED ISSUE #237: [Bug]: CI Pipelines currently fail after the introduction of FidelityFX
### Description

Currently we can't build artifacts because the pipelines are broken

### Steps to Reproduce

1. Try to release a build

### Expected Behavior

CI pipelines should pass

### Actual Behavior

CI pipelines fail

### Platform

Windows

### Clang version

_No response_

### Version / Commit

_No response_

_No response_

### Related

_No response_

sanjay900:
resolves #237

### DIFF
--- a/.github/workflows/build-linux-amd64.yaml
+++ b/.github/workflows/build-linux-amd64.yaml
@@ -37,9 +37,11 @@ jobs:
 
       - name: Install build dependencies
         run: |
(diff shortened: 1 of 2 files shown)


sanjay900:
Currently the windows build works fine, the linux build fails with a linker error, but i was able to replicate that locally with CI/CD, and found that specific issue is already fixed on  `dev/v0.5.0`

Issue #259, opened 2026-03-22
[Feature]: Trigger CVAR


SolarCookies:
### Summary

A way to call functions from the F4 Settings Menu

### Proposed Solution

Add a CVAR that has a `void (*fn)()` pointer that can be called using a button on the f4 menu

### Requirements

- [ ] Add the CVAR Variable so that it can be defined

### Out of Scope

Making trigger functions with inputs, Like Int where each time it changes it uses a callback `void (*fn)(int)`

### Acceptance Criteria

- [ ] Definable Trigger CVAR
- [ ] A ImGui button to call the trigger function

PR #260, opened 2026-03-22
Added Trigger Cvar

SolarCookies:
Adds a CVAR that stores a "void (*fn)()" function pointer. The F4 menu then calls the function using a ImGui Button. This is useful for actions like "Unlock All Items" rather then having to hook a tick function and check a BOOL CVAR we can use this trigger instead.

image

image

Implements: #259

### DIFF
--- a/include/rex/cvar.h
+++ b/include/rex/cvar.h
@@ -12,7 +12,7 @@
  * REXCVAR_DEFINE_STRING(my_string, "default", "Category", "A string setting");
  * @endcode
  *
- * Available types: BOOL, INT32, INT64, UINT32, UINT64, DOUBLE, STRING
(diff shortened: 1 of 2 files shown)


include/rex/cvar.h
```diff
@@ -141,6 +141,7 @@ struct FlagEntry {
   std::string description;
   std::function<bool(std::string_view)> setter;
   std::function<std::string()> getter;
+  void (*trigger_fn)() = nullptr;
```
twist84, line 144: Shouldn't this be using `std::function` like the rest of them.

SolarCookies: Thank you for the heads up I switched over to std::function in my new pull request #263

PR #274, opened 2026-04-04
Fixing Invalid crossover between backends in presenter.cpp when trying to build Vulkan on a windows machine (against the dev repo this time)

flashfire199:
Fixes vulkan on windows

### DIFF
--- a/src/ui/CMakeLists.txt
+++ b/src/ui/CMakeLists.txt
@@ -135,6 +135,7 @@ if(WIN32)
     target_link_libraries(rexui PUBLIC
         dwmapi
         Shcore
+        dxgi
     )
 else()
     find_package(PkgConfig REQUIRED)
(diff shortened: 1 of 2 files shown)


tomcl7:
Looks good to me. I would like @mrcmunir and @Mystixor to review and test before we push into development.

Looks fine to me. I would prefer if you squash these into a single commit with a message like `refactor(gpu): cleanup DX detection macros for compiling with VK backend` or something to that effect.

Nice work and Thank you!

mrcmunir (commented):
That seems correct tested it; I didn't see or detect any regression in that regard.

Mystixor (commented):
Tested on NG2-403. Everything compiles without issue on Windows, however the screen is permanently black. Even the debug overlays do not render when activating them. The game is otherwise running normal, as can be made out by pressing buttons and listening to the audio.

tomcl7:
I appreciate the attempt that you made with this PR, however I think it would best be handled with a more thorough review of the VK support on Windows.

Issue #279, opened 2026-04-11
[Bug]: fix(linux): build failures, audio crash, sampler assert, mouse sensitivity


etonedemid:
### Description

Five independent issues prevent clean builds and stable runtime on Linux x86_64.

1. Build fails without explicit -mssse3 / -msse4.1 flags on any machine not using -march=native.
   Memory, audio, and graphics code uses SSSE3/SSE4.1 intrinsics with no compile-time guard.

2. AudioSystem::Shutdown() calls DestroyDriver() unconditionally on any in-use client slot,
   crashing when a slot was reserved but never received a driver (e.g. when audio init failed).

3. AudioSystem::SubmitFrame() hits assert_true(driver != NULL), killing the process if audio
   is unavailable, rather than silently skipping frame submission.

4. SDLAudioDriver::Initialize() returns false immediately on SDL_INIT_AUDIO failure with no
   fallback, making the entire emulator refuse to start on headless or audio-less systems.

5. VulkanTextureCache::UseSampler() hits assert_false(mip_linear) when mip_base_map and
   mip_linear arrive together from conflicting fetch-constant state. Anisotropic filtering
   also does not clear mip_base_map, producing incorrect VkSampler parameters.

### Steps to Reproduce

1. Build with `cmake --preset linux-amd64 && cmake --build out/build/linux-amd64` on a
   machine without -march=native (e.g. inside a container or CI with -march=x86-64-v2)
2. Run the emulator on a system with no audio output device (e.g. headless server or WSL2)
3. Launch any title that uses anisotropic texture filtering
4. Observe: build error (1), immediate crash or silent hang (2-4), assert abort (5)

### Expected Behavior

- Build succeeds on any x86_64 Linux target without requiring -march=native
- Emulator starts and runs silently when no audio device is present (dummy driver fallback)
- No assertion failures or crashes from audio client slots lacking a driver
- VkSampler created correctly when aniso is active or mip_base_map + mip_linear conflict

### Actual Behavior

The program crashes with:
# Build error (issue 1):
error: always_inline function '_mm_shuffle_epi8' requires target feature 'ssse3'

# Runtime crash (issue 2-3):
Assertion failed: driver != NULL  [audio_system.cpp:263]

# Runtime abort (issue 5):
Assertion failed: !parameters.mip_linear  [texture_cache.cpp:874]

### Platform

Linux

Issue #278, opened 2026-04-11
[Feature]: feat(graphics): Xenos ZPD occlusion query support


etonedemid:
### Summary

Implement VulkanZPDQueryPool to enable guest occlusion query result readback.

The current Vulkan backend has no implementation for the Xbox 360 Xenos zero-pixel-draw
(ZPD) mechanism. Games that issue ZPASS_DONE queries receive no results, causing incorrect
conditional rendering, broken GPU-driven culling, and in some titles visual corruption or
hangs.

### Proposed Solution

Build on VK_QUERY_TYPE_OCCLUSION with a VulkanZPDQueryPool managing per-frame segment lifecycle.

### Key Features
- **VulkanZPDQueryPool**: VK_QUERY_TYPE_OCCLUSION pool with streaming result readback, segment open/close, and capacity growth
- **XenosZPDReport layout**: guest-memory report struct mirroring the Xenos GPU register format (32-byte records, 64-byte slots)
- **DeferredCommandBuffer**: accumulates ZPD commands before flushing into a primary command buffer
- **CP integration**: OpenQuerySegment / CloseQuerySegment injected around render pass enter/leave and at EndSubmission
- **SPIR-V crash fixes**: three null-ID crashes in the fragment shader color-target output path fixed as a prerequisite

### Occlusion query pool
- [ ] VulkanZPDQueryPool creates VK_QUERY_TYPE_OCCLUSION pool via vkCreateQueryPool
- [ ] Pool resets query slots with vkResetQueryPool (EXT_host_query_reset)
- [ ] OpenQuerySegment / CloseQuerySegment wrap each draw segment
- [ ] Results are read back and written into guest memory in XenosZPDReport layout

### Command processor integration
- [ ] OpenQuerySegment(false) called after in_render_pass_ = true in both SubmitBarriersAndEnterRenderTargetCacheRenderPass overloads
- [ ] CloseQuerySegment() called in EndRenderPass() with pending-begin logic
- [ ] CloseQuerySegment() called before SubmitBarriers(true) in EndSubmission

### SPIR-V prerequisite fixes
- [ ] Non-FSI color-target loop masks color_targets_remaining by color_targets_used
- [ ] RT0 alpha-test / a2c paths use StorageClassFunction fallback when RT0 not in color_targets_used
- [ ] Fallback variable xe_var_fragment_data_0_rt_masked created in StartFragmentShaderInMain() body scope

### Out of Scope

[...]

PR #283, opened 2026-04-12
Fix GPU issues, mouse capture, vehicle remapping, and gamepad navigation

etonedemid:
## Summary

Four independent fixes bundled together: corrects codegen precision bugs for FMA/frsqrte, fixes several GPU rendering issues on Vulkan, adds proper mouse capture on Wayland/X11, introduces memory-mapped vehicle input remapping, and makes ImGui gamepad navigation functional.

---

### Codegen — FMA / frsqrte precision (`src/codegen/builders/floating_point.cpp`)

- **`fmadd` / `fmsub` / `fnmadd` / `fnmsub`**: replace `a * b ± c` with `std::fma(a, b, ±c)`. The old expansion was subject to double-rounding: the compiler could fold the multiply-add into a hardware FMA or not, giving non-deterministic results inconsistent with the Xbox 360's non-fused semantics. `std::fma` guarantees a single correctly-rounded result.
- **`frsqrte`**: narrow the result through `float` before returning as `double` (`double(float(1.0 / sqrt(x)))`), matching the Xbox 360's reduced-precision reciprocal square-root estimate.
- Both single- and double-precision variants updated for all four FMA forms.

### Vulkan Render Target Cache (`src/graphics/vulkan/render_target_cache.cpp`)

- **K8_A8B8G8R8 format**: correct Vulkan format mapping from `VK_FORMAT_A8B8G8R8_UNORM_PACK32` → `VK_FORMAT_A2B10G10R10_UNORM_PACK32`.
- **SPIRV version**: utility shader `SpirvBuilder` instances now use the device's actual SPIRV version instead of the hardcoded `spv::Spv_1_0`, preventing validation errors on drivers that require 1.3+.
- **Resolve failure spam**: rate-limit `GetResolveInfo` failure messages (log first 5, then every 1000) to prevent log flooding during normal gameplay.
- **D24 fallback warning**: emit a one-time warning when `D24_UNORM_S8_UINT` is unavailable and the driver falls back to `D32_SFLOAT_S8_UINT`.

### Vulkan Command Processor (`src/graphics/vulkan/command_processor.cpp`)

- **`EVENT_WRITE_ZPD` / occlusion queries**: add `occlusion_query` CVAR (`fake` / `fast` / `strict`) controlling how hardware visibility queries are handled. `fake` writes a synthetic sample count (range configurable via `occlusion_query_fake_lower/upper_threshold` CVARs); `fast` queries the GPU without stalling; `strict` waits for the real result. Resolves hangs and assertion failures in titles that rely on ZPD feedback.

### Vulkan Pipeline Cache (`src/graphics/vulkan/pipeline_cache.cpp`)

- Additional FMA/frsqrte lowering fixes mirroring the codegen changes, applied at the SPIR-V translation layer.

### Mouse Capture — Wayland + X11 (`src/ui/window_gtk.cpp`, `include/rex/ui/window_gtk.h`)

- **Wayland**: call `gdk_seat_grab()` / `gdk_seat_ungrab()` when toggling mouse capture so the cursor is actually confined to the window (Wayland does not support `XWarpPointer`).
- **X11**: guard `XWarpPointer` behind a compile-time X11 macro to eliminate spurious warnings on Wayland-only builds.

### Mouse Input — CenterCursor race fix (`src/input/mnk/mnk_input_driver.cpp`)

- Defer warp requests via an atomic flag rather than calling `XWarpPointer` directly from the UI thread; execute on the input thread to eliminate the race between cursor centering and motion event delivery.

### Vehicle Mode Input Remap (`include/rex/system/kernel_state.h`, `src/system/kernel_state.cpp`, `include/rex/input/input_driver.h`, `include/rex/input/input_system.h`, `src/input/input_system.cpp`, `src/input/mnk/mnk_input_driver.cpp`, `src/ui/rex_app.cpp`)

- Add four CVARs (`vehicle_steering_addr`, `vehicle_throttle_addr`, `vehicle_brake_addr`, `vehicle_handbrake_addr`) for specifying guest memory addresses of analog control values.
- Add `MemoryReadU32Fn` callback type to `InputDriver`; wire a `memory->TranslateVirtual<uint32_t*>` lambda from `ReXApp::OnInitialize` through `InputSystem` to `MnkInputDriver`.
- `TranslateVirtual` reads the mapped address and byte-swaps the value to drive the alt-mode vehicle axis, so the correct physical input feeds the control without needing title-specific patches.

### ImGui Gamepad Navigation (`include/rex/ui/imgui_drawer.h`, `src/ui/imgui_drawer.cpp`)

- Enable `ImGuiConfigFlags_NavEnableGamepad` in `Initialize()`.
- Poll the first available SDL gamepad each frame before `ImGui::NewFrame()` using `SDL_OpenGamepad()` (ref-counted — safe even before `SDLInputDriver` processes `SDL_EVENT_GAMEPAD_ADDED`). Release with `SDL_CloseGamepad()` each frame; the input driver retains its own reference.
- Full button mapping: D-pad, A/B/X/Y, Start/Back, LB/RB, LT/RT (analog), left stick (four analog directions).
- `ImGuiBackendFlags_HasGamepad` set/cleared based on gamepad availability.

---

## Testing

- Built and linked against rexglue-sdk RelWithDebInfo on Linux x86-64 (Clang).
- FMA/frsqrte changes verified against PPC test suite (existing codegen tests pass).

#278 #279


mrcmunir:
needed re-run clang-format -i in those files for correctly clang format.

etonedemid:
> needed re-run clang-format -i in those files for correctly clang format. <

@mrcmunir 
done

tomcl7 (changes requested):
Before I go too far into reviewing this, can you rewrite your commit history to reflect the changes more granularly? The way you have the changes laid out in the PR is a nice and logical way to rewrite the commits.

Also rebase against development for this PR.

src/codegen/builders/floating_point.cpp
tomcl7, line 1: Drop 374aa140678e2b4c6f986f02fb99f23ea1d3a4b1 it has been cherry picked.

src/graphics/vulkan/render_target_cache.cpp
```diff
@@ -203,6 +203,9 @@ bool VulkanRenderTargetCache::Initialize(uint32_t shared_memory_binding_count) {
   const VkDevice device = vulkan_device->device();
   const ui::vulkan::VulkanDevice::Properties& device_properties = vulkan_device->properties();
 
+  // Cache the SPIR-V version for utility shader creation.
+  spirv_version_ = SpirvShaderTranslator::Features(vulkan_device).spirv_version;
```
tomcl7, line 207: `spirv_version_` isn't declared anywhere on this branch. This produces a compile error against upstream `development`. Could you add the member declaration as part of this commit, or pull in whatever base commit defines it?

src/graphics/vulkan/render_target_cache.cpp
```diff
@@ -1341,9 +1349,17 @@ bool VulkanRenderTargetCache::Resolve(const memory::Memory& memory,
...
+    static uint32_t resolve_fail_count = 0;
+    ++resolve_fail_count;
+    if (resolve_fail_count <= 5 || (resolve_fail_count % 1000 == 0)) {
+      REXGPU_ERROR("VulkanRenderTargetCache::Resolve: GetResolveInfo failed (occurrence #{})",
+                   resolve_fail_count);
+    }
     return false;
   }
 
+  // Resolve logging removed — was spamming.
```
tomcl7: Please drop this comment. There was no logging there to begin with?

src/graphics/vulkan/render_target_cache.cpp
```diff
@@ -2513,7 +2529,8 @@ VkShaderModule VulkanRenderTargetCache::GetTransferShader(TransferShaderKey key)
   std::vector<spv::Id> id_vector_temp;
   std::vector<unsigned int> uint_vector_temp;
 
-  SpirvBuilder builder(spv::Spv_1_0, (SpirvShaderTranslator::kSpirvMagicToolId << 16) | 1, nullptr);
+  SpirvBuilder builder(static_cast<spv::SpvVersion>(spirv_version_),
```
tomcl7, line 2530: Same compile error as line 207.

src/graphics/vulkan/render_target_cache.cpp
```diff
@@ -5619,7 +5636,8 @@ VkPipeline VulkanRenderTargetCache::GetDumpPipeline(DumpPipelineKey key) {
 
   std::vector<spv::Id> id_vector_temp;
 
-  SpirvBuilder builder(spv::Spv_1_0, (SpirvShaderTranslator::kSpirvMagicToolId << 16) | 1, nullptr);
+  SpirvBuilder builder(static_cast<spv::SpvVersion>(spirv_version_),
```
tomcl7, line 5637: Same compile error as line 207.

src/graphics/vulkan/pipeline_cache.cpp
```diff
@@ -291,9 +292,22 @@ bool VulkanPipelineCache::Initialize() {
...
+
+  // When FSI is active, suppress demote_to_helper_invocation.
+  // The FSI alpha test path via OpDemoteToHelperInvocationEXT + OpIsHelperInvocationEXT
+  // is unreliable on NVIDIA (SPIR-V 1.5): demoted invocations may not be properly
+  // flagged, causing alpha-killed pixels to write their output colour to EDRAM.
+  // The non-DemoteToHelper path directly zeroes fsi_sample_mask_in_rt_0_alpha_tests,
+  // which is correct and avoids the driver query entirely.
+  if (edram_fragment_shader_interlock) {
+    translator_features.demote_to_helper_invocation = false;
+  }
```
tomcl7, line 304: This is a piece I'd like to take, but it's wedged in alongside the broken changes below. Could you extract just this FSI workaround into a standalone commit (e.g. `fix(gpu/vulkan): disable demote_to_helper_invocation on FSI path for NVIDIA`)? It doesn't depend on anything else and would land immediately.

src/graphics/vulkan/pipeline_cache.cpp
```diff
@@ -960,7 +974,7 @@ SpirvShaderTranslator::Modification VulkanPipelineCache::GetCurrentVertexShaderM
 
 SpirvShaderTranslator::Modification VulkanPipelineCache::GetCurrentPixelShaderModification(
     const Shader& shader, uint32_t interpolator_mask, uint32_t param_gen_pos,
-    reg::RB_DEPTHCONTROL normalized_depth_control) const {
+    uint32_t normalized_color_mask, reg::RB_DEPTHCONTROL normalized_depth_control) const {
```
tomcl7: New parameter on `GetCurrentPixelShaderModification`, but `include/rex/graphics/vulkan/pipeline_cache.h:` still declares the 4-arg version, and the only caller still passes 4 args. Header and caller need the matching update in the same commit or it won't build.

src/graphics/vulkan/pipeline_cache.cpp
```diff
@@ -1005,6 +1019,12 @@ SpirvShaderTranslator::Modification VulkanPipelineCache::GetCurrentPixelShaderMo
         modification.pixel.depth_stencil_mode = DepthStencilMode::kNoModifiers;
       }
     }
+
+    // Extract 1 bit per RT from the 4-bits-per-RT normalized_color_mask.
+    modification.pixel.color_targets_used = (((normalized_color_mask >> 0) & 0xF) ? 1 : 0) |
+                                            (((normalized_color_mask >> 4) & 0xF) ? 2 : 0) |
+                                            (((normalized_color_mask >> 8) & 0xF) ? 4 : 0) |
+                                            (((normalized_color_mask >> 12) & 0xF) ? 8 : 0);
```
tomcl7: `Modification::pixel` doesn't have a `color_targets_used` field on this branch. Compile error.

src/input/CMakeLists.txt
```diff
@@ -19,3 +19,11 @@ else()
 endif()
 
 target_link_libraries(rexinput PUBLIC rexcore SDL3::SDL3)
+
+if(NOT WIN32)
+    find_package(PkgConfig REQUIRED)
+    pkg_check_modules(GTK3 REQUIRED gtk+-3.0)
+    pkg_check_modules(X11 REQUIRED x11)
```
tomcl7: Two things on this line:

1. Wrong commit -  The X11 link is for 6c724a4aa2e7a2a2511087af0c1927065ccdaa95 `XWarpPointer` call this commit (3bf16479626c21ad10144fd0cc60b1b669b14eef) doesn't use X11, only GTK/GDK. Could you move this line into 6c724a4aa2e7a2a2511087af0c1927065ccdaa95 where it belongs? Makes the history accurate and the dependency graph cleaner.
2. `REQUIRED` on x11 on all non-Windows platforms will make pure Wayland distros without `libx11-dev` fail to configure. Could you gate this behind a CMake option (e.g. `REX_INPUT_ENABLE_X11` default ON) so it can be opted out?

src/ui/rex_app.cpp
```diff
@@ -232,33 +233,11 @@ bool ReXApp::OnInitialize() {
...
-          } else {
-            settings_overlay_ =
-                std::make_unique<ui::SettingsDialog>(imgui_drawer_.get(), config_path_);
-          }
-        });
+        // Built-in overlays
+        debug_overlay_ = std::make_unique<ui::DebugOverlayDialog>(imgui_drawer_.get());
+        console_overlay_ = std::make_unique<ui::ConsoleDialog>(imgui_drawer_.get(), log_sink_);
+        settings_overlay_ = std::make_unique<ui::SettingsDialog>(
+            imgui_drawer_.get(), exe_dir / (std::string(GetName()) + ".toml"));
```
tomcl7: **Please don't remove these keybinds.** They're intentional.

> Overlay keybinds - dialogs are created/destroyed on demand so ImGuiDrawer can detach when idle, enabling kGuestOutputThreadImmediately paint mode for 1:1 host-guest frame sync.

Always-constructing the overlays defeats that optimization permanently, even during normal gameplay when no overlay is visible. ImGuiDrawer can never detach, so the immediate-paint fast path is permanently off. If the problem you hit was "a keybind wasn't working", let's investigate that root cause instead.

src/ui/rex_app.cpp
```diff
@@ -269,17 +248,24 @@ bool ReXApp::OnInitialize() {
...
+  // This loads previously compiled pipeline descriptions from disk and
+  // pre-compiles them asynchronously, avoiding shader compilation stalls
+  // on subsequent runs.
+  if (auto* gs = dynamic_cast<rex::graphics::GraphicsSystem*>(runtime_->graphics_system())) {
+    auto title_id = runtime_->kernel_state()->title_id();
+    if (title_id && !user_data_root_.empty()) {
+      gs->InitializeShaderStorage(user_data_root_, title_id, false);
+    }
+  }
+
```
tomcl7: `InitializeShaderStorage` is already wired up at line 287 of the same file using `runtime_->cache_root()` in blocking mode, after the title ID is established. 

The commit title says "remove cache_root" but `cache_root()` is still referenced at line 287. Additionally, `cache_root_` is a purpose-built path distinct from user_data_root (cvar-overridable via `cache_path`, defaults to `user_dir/cache`). This new earlier call also runs before the deferred module launch, where the title ID may not yet be populated.

src/ui/rex_app.cpp
```diff
@@ -269,17 +248,24 @@ bool ReXApp::OnInitialize() {
         // (e.g. overlay is open). This controls MnK mouse capture.
         auto* input_sys = static_cast<rex::input::InputSystem*>(runtime_->input_system());
         if (input_sys) {
-          input_sys->SetActiveCallback([this]() {
-            if (!debug_overlay_ && !console_overlay_ && !settings_overlay_)
-              return true;
-            return !ImGui::GetIO().WantCaptureMouse;
-          });
+          input_sys->SetActiveCallback([]() { return !ImGui::GetIO().WantCaptureMouse; });
```
tomcl7: Related to the overlay removal above, the old form short-circuited `true` when no overlay existed (the common case with on-demand overlays). With always-constructed overlays that short-circuit is never hit. Behavior diverges whenever ImGui's capture flag flips.

src/system/xmemory.cpp
```diff
@@ -180,11 +180,22 @@ bool Memory::Initialize() {
                               0x1FD00000, 4096, &heaps_.physical);
 
   // Protect the first and last 64kb of memory.
+  // Always commit as read-write first so we can scribble a non-zero pattern,
+  // then downgrade protection.  Games that dereference null pointers will read
+  // 0xCD instead of 0x00, which prevents silent data-dependent rendering bugs
+  // (the "black screen" issue on Linux where zeroed null-page memory causes
+  // the game to skip geometry submission).
```
tomcl7: This silently repurposes the protect_zero cvar. 

Before, protect_zero=false meant Read|Write, null reads returned 0x00 and null writes silently succeeded. 
After, it scribbles 0xCD and sets ReadOnly, so reads return 0xCD and writes that used to succeed now fault. Same cvar name, new third mode, no way to get the old permissive behavior back without a code change. Anyone running protect_zero=false as a workaround will silently get different crashes. Also, in strict mode (protect_zero=true) the memset runs 64KB and the page is immediately NoAccess. Dead work.

If this is worth implementing, please gate it on a new cvar (e.g. null_page_fill=zero|cd|off), only memset when the page stays readable, and name the title in the commit message. Otherwise please drop this; the other four fixes in df36c121bb076d446a3445893dd3fdf315518687 are fine.

tomcl7 (changes requested):
- Title for 71acf85261ad8ba5a6cb5ba0dfb5c2b19b997919 is inaccurate. The title says "K8A8B8G8R8" which is neither a Xenos nor a Vulkan identifier. The actual fix is the `k_2_10_10_10` --> `A2B10G10R10_UNORM_PACK32` mapping. Could you reword the title so it's searchable later?
- Title for 302e854f327258741644686c0cd5816701856660 is inaccurate. There's no FMA or frsqrte work in here. Could you fix the title? It makes the commit very hard to find later when someone goes looking for the FSI workaround.
- 4760801f5efdff0bdaf97d128d61b747ab8440e4 references a substantial amount of ZPD occlusion query infrastructure that isn't on this branch: `XenosZPDReport::*`, `zpd_active_segment_`, `zpd_batch_fake_`, and quite a few others. None of these are declared in this tree. This isn't a splittable situation, the whole thing is built on top of that infrastructure. Could you open a separate PR that lands it as a standalone foundation with this work?
- Drop 37964f76b15f62cce2ba0ba1fd9bc664fd9dd050 entirely

src/graphics/vulkan/render_target_cache.cpp
etonedemid: Sorry, had local logging

etonedemid:
something is seriously wrong

PR #287, opened 2026-04-13
feat(krnl): Add stubs for xbdm

### LINKED ISSUE #191: [Feature]: xbdm ordinal support
### Summary

Running codegen on Dead Space (Apr 20, 2007 prototype) produces a handful of xbdm ordinals that cannot be resolved:

`[error] [codegen] [t28568] Cannot resolve ordinal 4 from xbdm`
`[error] [codegen] [t28568] Cannot resolve ordinal 51 from xbdm`
`[error] [codegen] [t28568] Cannot resolve ordinal 82 from xbdm`
`[error] [codegen] [t28568] Cannot resolve ordinal 141 from xbdm`

Popping this one as a feature since technically codegen can run just fine without it even if it displays as an error
(Background: config .toml was produced by rexglue's IDA helper script with functions defined by the accompanying .pdb)

### Proposed Solution

v0.3.0 of the SDK already adds ordinal resolution so this should be hypothetically within scope and able to be implemented with the same foundation

### Requirements

In this case Im not sure if there is any real amount of xbdm ordinals in existence above the amount that are showing up in my current codegen attempts so this might need more looking into when adjusting the rexglue kernel for this issue (since naturally this issue requires more kernel adjustments)

### Out of Scope

There is still an unresolved call in this case with the IDA helper script + .pdb which is certainly a curiosity of its own but it isn't really related to the xbdm stuff

### Modified Dependencies

_No response_

### Acceptance Criteria

I assume it should be simple enough to see that it works once it resolves through codegen

### Related

_No response_

josephl70:
## Summary
Adds stubs for the xbdm functions, and fixes the ordinal errors relating to xbdm
Should fix #191


include/rex/kernel/xbdm/module_export_groups.inc
```diff
@@ -0,0 +1,13 @@
...
+ ******************************************************************************
+ * Copyright 2022 Ben Vanik. All rights reserved.                             *
+ * Released under the BSD license - see LICENSE in the root for more details. *
+ ******************************************************************************
+ */
+
+// This is a partial file designed to be included by other files when
+// constructing various tables.
+
+XE_MODULE_EXPORT_GROUP(xbdm, Misc)
```
tomcl7: Missing trailing newline

src/kernel/xbdm/xbdm_misc.cpp
```diff
@@ -0,0 +1,270 @@
...
+REX_EXPORT_STUB(__imp__XLFInitializeUserLog);
+REX_EXPORT_STUB(__imp__XLFEndLog);
+REX_EXPORT_STUB(__imp__XLFLogPrint);
+REX_EXPORT_STUB(__imp__XLFLogPrintV);
+REX_EXPORT_STUB(__imp__XLFLogBuffer);
+REX_EXPORT_STUB(__imp__XLFLogMessageStats);
+REX_EXPORT_STUB(__imp__XLockFreeGetErrorHandler);
+REX_EXPORT_STUB(__imp__XLockFreeSetErrorHandler);
+REX_EXPORT_STUB(__imp__DmExecuteThreadRPC);
+REX_EXPORT_STUB(__imp__DmGetDebuggerConnection);
```
tomcl7: Missing trailing newline

tomcl7 (changes requested):
Very nicely done. Thank you for getting this stub scaffold setup.

Only minor thing is 2 files are missing a trailing newline.

josephl70:
Fixed :D

PR #289, opened 2026-04-14
codegen: vpkd3d128 tweak

Graine25:


src/codegen/builders/vector.cpp
```diff
@@ -1319,17 +1319,32 @@ bool build_vpkd3d128(BuilderContext& ctx) {
       break;
     }
 
-    case 5:  // float16_4
+    case 5:  // float16_4 - Fix by mzzvxm
```
tomcl7: Drop the "Fix by mzzvxm" tag from the source comment. Authorship belongs in the commit/co-author trailer, not in the code.

src/codegen/builders/vector.cpp
```diff
@@ -1319,17 +1319,32 @@ bool build_vpkd3d128(BuilderContext& ctx) {
...
     {
-      // Pack 4 elements into 64 bits (4 x 16-bit floats)
-      // Guest element 0 goes to highest 16-bit position, element 3 to lowest
-      // Output u16 index = (3-i) + 2*shift for element i
-      if (ctx.insn.operands[3] != 2 || ctx.insn.operands[4] > 2)
+      // Combinations observed in the game:
+      //   mask=2, shift=0 → write u16[3..0], zero u16[4..7]
+      //   mask=2, shift=2 → write u16[7..4], zero u16[3..0] (first pass)
+      //   mask=3, shift=0 → write u16[3..0] without zeroing (second pass, preserves the upper half)
+      if (ctx.insn.operands[4] > 3)
```
tomcl7: Two issues with this guard:

1. The shift bound is too loose. `dstIdx = (3 - i) + 2 * shift;` with `shift == 3` that ranges `6..9`, which overruns the vector's 8 `u16` lanes (valid indices are `0..7`). The previous > 2 bound actually protected against this. Restore it or gate the loop on `dstIdx < 8` so out-of-range cases hit the `REXCODEGEN_WARN` path instead of silently emitting OOB stores.
2. Unexpected mask values (0, 1) now fall through silently with no zeroing and no diagnostic. Please also warn when `mask != 2 && mask != 3,` so unseen encodings get flagged instead of miscompiled.

src/codegen/builders/vector.cpp
```diff
@@ -1319,17 +1319,32 @@ bool build_vpkd3d128(BuilderContext& ctx) {
...
+    case 5:  // float16_4 - Fix by mzzvxm
     {
-      // Pack 4 elements into 64 bits (4 x 16-bit floats)
-      // Guest element 0 goes to highest 16-bit position, element 3 to lowest
-      // Output u16 index = (3-i) + 2*shift for element i
-      if (ctx.insn.operands[3] != 2 || ctx.insn.operands[4] > 2)
+      // Combinations observed in the game:
+      //   mask=2, shift=0 → write u16[3..0], zero u16[4..7]
+      //   mask=2, shift=2 → write u16[7..4], zero u16[3..0] (first pass)
+      //   mask=3, shift=0 → write u16[3..0] without zeroing (second pass, preserves the upper half)
```
tomcl7: Worth calling out that these combos come from game traces, not the [spec](https://github.com/rexglue/rexglue-sdk/blob/main/docs/ppc/vmx128.txt). The [spec ](https://github.com/rexglue/rexglue-sdk/blob/main/docs/ppc/vmx128.txt)only defines the encoding, not what `y` and `z` actually mean, so anyone reading this later should know the handling is heuristic and may need extending if new combos turn up.

src/codegen/builders/vector.cpp
```diff
@@ -1319,17 +1319,32 @@ bool build_vpkd3d128(BuilderContext& ctx) {
...
 
+      uint32_t mask = ctx.insn.operands[3];
+      uint32_t shift = ctx.insn.operands[4];
+
+      // mask=2: before writing, clear the half that will NOT be written
+      // (ensures the opposite half has no garbage when this is the only step)
+      if (mask == 2) {
+        // shift=0 → write u16[0..3], zero u16[4..7] (u32[2] and u32[3])
+        // shift=2 → write u16[4..7], zero u16[0..3] (u32[0] and u32[1])
+        size_t clearU32Start = (shift == 0) ? 2 : 0;
```
tomcl7: This only handles `shift in {0, 2}`. For `shift == 1` (still allowed by the new bound), this picks `clearU32Start = 0` and zeroes `u32[0..1]`, but the subsequent writes target `u16[2..5]`, so the upper half `u16[6..7]` never gets cleared and the lower clear partially overlaps the written lanes. Either restrict to `shift` in {0, 2} explicitly, or make the clear derive from the `dstIdx` range.

src/codegen/builders/vector.cpp
```diff
@@ -1319,17 +1319,32 @@ bool build_vpkd3d128(BuilderContext& ctx) {
...
+        ctx.println("\t{}.u32[{}] = 0;", ctx.v(ctx.insn.operands[0]), clearU32Start);
+        ctx.println("\t{}.u32[{}] = 0;", ctx.v(ctx.insn.operands[0]), clearU32Start + 1);
```
tomcl7: Consider emitting a single `u64[n] = 0;` for the 64-bit half. One line of generated code instead of two, and matches the "clear the opposite half" intent more directly.

src/codegen/builders/vector.cpp
```diff
@@ -1319,17 +1319,32 @@ bool build_vpkd3d128(BuilderContext& ctx) {
...
+      }
+
       for (size_t i = 0; i < 4; i++) {
-        size_t srcIdx = 3 - i;  // Guest element i is at host array index 3-i
-        size_t dstIdx = (3 - i) + (2 * ctx.insn.operands[4]);  // Output also reversed
+        size_t srcIdx = 3 - i;
+        size_t dstIdx = (3 - i) + (2 * shift);
```
tomcl7, line 1362: Tie this to the bounds-check fix above. Once `shift` is capped at 2, this is safe. Worth leaving a short note about the invariant (`dstIdx` must stay under 8) so the next person touching this does not widen the bound again.

tomcl7 (changes requested):
Fix the commit message. "Tweak" undersells the change. This broadens accepted operand combinations and changes emitted code for existing ones. Prefer something like `feat(codegen): support mask=3 and mask=2 zero-clear for vpkd3d128 float16_4`, with a one-line note about the observed game traces.

Overall, the direction is right. The old guard was too strict and blocked a real combo, but the new guard is too permissive in two directions (`shift==3` OOB, unvalidated `mask`), and the `mask==2` clear logic does not cover all currently allowed shifts. Tightening the bounds and fixing the clear's shift handling would make this mergeable.

src/codegen/builders/vector.cpp
```diff
@@ -1321,15 +1321,44 @@ bool build_vpkd3d128(BuilderContext& ctx) {
...
+
+      // Guard fix: The shift bound was too loose (shift=3 would cause dstIdx to reach 9, an OOB
+      // store). Also, we explicitly warn on unexpected mask values (e.g., 0, 1) to avoid silent
+      // fallthroughs and silently generating miscompiled code.
+      if (shift > 2 || (mask != 2 && mask != 3)) {
+        REXCODEGEN_WARN("Unexpected float16_4 pack instruction at {:X} (mask={}, shift={})",
+                        ctx.base, mask, shift);
+        // Optional: emit a debug trap in the generated code to catch this at runtime
+        ctx.println("\t__builtin_debugtrap();");
+        return true;  // or false, depending on RexGlue error handling architecture
```
tomcl7: Drop the hedge. Other traps in this file (1403, 1548) return true; do the same and delete the comment.

src/codegen/builders/vector.cpp
```diff
@@ -1321,15 +1321,44 @@ bool build_vpkd3d128(BuilderContext& ctx) {
...
+      uint32_t shift = ctx.insn.operands[4];
+
+      // Guard fix: The shift bound was too loose (shift=3 would cause dstIdx to reach 9, an OOB
+      // store). Also, we explicitly warn on unexpected mask values (e.g., 0, 1) to avoid silent
+      // fallthroughs and silently generating miscompiled code.
+      if (shift > 2 || (mask != 2 && mask != 3)) {
+        REXCODEGEN_WARN("Unexpected float16_4 pack instruction at {:X} (mask={}, shift={})",
+                        ctx.base, mask, shift);
+        // Optional: emit a debug trap in the generated code to catch this at runtime
+        ctx.println("\t__builtin_debugtrap();");
```
tomcl7: It is not optional, it is there. State what it does, not that it was considered.

src/codegen/builders/vector.cpp
```diff
@@ -1321,15 +1321,44 @@ bool build_vpkd3d128(BuilderContext& ctx) {
...
+      // mask=2, shift=2 → write u16[7..4], zero u16[3..0] (first pass)
+      // mask=3, shift=0 → write u16[3..0] without zeroing (second pass, preserves the upper half)
+
+      uint32_t mask = ctx.insn.operands[3];
+      uint32_t shift = ctx.insn.operands[4];
+
+      // Guard fix: The shift bound was too loose (shift=3 would cause dstIdx to reach 9, an OOB
+      // store). Also, we explicitly warn on unexpected mask values (e.g., 0, 1) to avoid silent
+      // fallthroughs and silently generating miscompiled code.
+      if (shift > 2 || (mask != 2 && mask != 3)) {
```
tomcl7: Guard still allows `shift == 1`, but the clear block on line 1347 only runs for shift in `{0, 2}`. So `mask == 2`, `shift == 1` emits writes to `u16[2..5]` with no clear on the opposite half. Either reject `shift == 1` here, or document why no clear is needed.

tomcl7 (commented):
Close, still a few things to fix.

tomcl7 (approved):

PR #295, opened 2026-04-25
refactor(system): migrate from legacy move_only functions

Graine25:
Pulled from xenia-edge, mainly for weird systems who don't supply this function automatically cough mac
https://github.com/has207/xenia-edge/blob/edge/src/xenia/base/threading_timer_queue.h/#L28

### DIFF
--- a/include/rex/system/kernel_state.h
+++ b/include/rex/system/kernel_state.h
@@ -285,24 +285,23 @@ class KernelState {
   void CompleteOverlappedImmediateEx(uint32_t overlapped_ptr, X_RESULT result,
                                      uint32_t extended_error, uint32_t length);
 
-  void CompleteOverlappedDeferred(std::move_only_function<void()> completion_callback,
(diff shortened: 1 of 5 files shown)


tomcl7 (approved):
Nice clean mechanical swap.

  - The pre_callback/post_callback are still invoked exactly once (in the dispatch lambda), so the extra copies during forwarding are stack-local and destroyed on unwind.
  - Dropping mutable is fine because `std::function::operator()` is already const-correct and still forwards into mutable wrapped callables.
  - The `sequence_t` --> `size_t` cast is guarded by an earlier `diff >= 0` check and matches the parallel try_claim overload.
  - No current caller passes a move-only callable, so relaxing to std::function breaks nothing.

The only thing that's technically taking a hit here is performance. `std::function` may heap allocate where `move_only_function` wouldn't, and forwarding copies instead of moves is wasteful. But for kernel overlapped completion and timer scheduling that's negligible, and it's the tradeoff for macOS portability.

PR #296, opened 2026-04-26
feat(krnl): Improved xbdm export stubs

twist84:
- Change **DmRegisterCommandProcessorEx** & **DmWalkLoadedModules** result codes

Previously Halo 3 debug would freeze if **DmRegisterCommandProcessorEx** didn't return success, and would get stuck in an endless loop if **DmWalkLoadedModules** didn't return 0x82DA0104, updated both to return what the game seemed to expect.

Affected titles:
- Halo 3 'delta' March 2007 - debug build can now start up

This is a change @emoose made for Xenia in the following [pull request](https://github.com/xenia-project/xenia/pull/1731)


tomcl7 (approved):
Well done. Thank you.

PR #297, opened 2026-04-26
Fix: always mask SSE exceptions in FPSCRPlatform::setcsr (Linux SIGFPE)

birabittoh:
## Problem

`InitHostExceptions` sets SSE exception mask bits at startup, but PPC
FPSCR values written at runtime translate directly to `ldmxcsr` with the
guest value, which can clear those mask bits and re-enable SSE exceptions.
On Linux this causes a `SIGFPE`. Windows is unaffected because SEH handles
floating-point exceptions transparently.

`setcsr` is called on every PPC FPSCR write from recompiled guest code, so
the exception mask is clobbered at high frequency during normal execution.

## Fix

OR `ExceptionMask` unconditionally in `setcsr` (x86_64 branch only), so
SSE exceptions remain masked for the lifetime of the process regardless of
what the PPC guest writes to FPSCR.

The ARM64 branch does not need this change. `InitHostExceptions` clears
exception enable bits there (ARM FPCR: 0 = disabled), and PPC FPSCR writes
do not map to ARM exception-enable bits in the same way.

### DIFF
--- a/include/rex/platform/fpscr.h
+++ b/include/rex/platform/fpscr.h
@@ -43,7 +43,7 @@ struct FPSCRPlatform {
 
   static inline u32 getcsr() noexcept { return simde_mm_getcsr(); }
 
-  static inline void setcsr(u32 csr) noexcept { simde_mm_setcsr(csr); }
+  static inline void setcsr(u32 csr) noexcept { simde_mm_setcsr(csr | ExceptionMask); }
 
   static inline void InitHostExceptions(u32& csr) noexcept {
     csr |= ExceptionMask;  // Set mask bits to disable exceptions


tomcl7:
Thanks for looking into this. The Linux/Windows symptom and the SEH vs SIGFPE explanation match what I'd expect. I have some concerns about the diagnosis and the fix itself though.

The PR description says guest FPSCR writes "translate directly to `ldmxcsr` with the guest value", which is what motivates putting the OR in `setcsr`. Reading the path though, I'm not seeing that. `storeFromGuest` only ever rewrites the rounding bits in the cached `csr`:

```cpp
csr &= ~RoundMaskVal;
csr |= Platform::GuestToHost[value & kRoundMask];
setcsr(csr);
```

The flush helpers only touch `FlushMask`. Every `setcsr` call goes through the cached member, never the raw guest value. So as long as `InitHost()` sets `ExceptionMask` once, those bits should ride through every `mtfsf` and flush flip untouched. I can't see a path where high-frequency guest FPSCR traffic would clear them.

On the ARM point specifically, I don't think the reasoning in the description for skipping it holds. The argument is that ARM uses enable bits (0 = disabled) instead of mask bits, so PPC FPSCR writes "do not map to ARM exception-enable bits in the same way". But the cached `csr` and the save/restore path are arch-agnostic. If the premise is that something is leaking guest bits into the cache and then into the host CSR, ARM is exposed to the same problem, just at different bit positions (IOE/DZE/OFE/UFE/IXE/IDE at bits 8-12,15). The bit-polarity difference is not a defense, it just means the failure mode on ARM is "exceptions get enabled when guest bits happen to land in those slots" instead of "masks get cleared". Either we fix the underlying invariant for both arches, or we accept that the diagnosis doesn't actually require a fix at all.

A few other things about the fix that concern me:

- It taxes every FPSCR write in recompiled code, which is hot, to defend an invariant that should already hold.
- It quietly removes the option of ever using `setcsr` to enable exceptions on purpose (e.g., debugging NaN sources in a dev build).
- If the cache really is being corrupted somewhere, this masks it rather than surfacing it.

This really needs a formal bug report before we change anything in `setcsr`. We have a bug report template in the repo, please file one with the failing trace and the steps to reproduce. I'd like to see which path is actually feeding bad bits to the host before we decide where the guard belongs.

Feel free to re-submit once an issue is opened with specific details and traces/debug info to back up the claims.

PR #300, opened 2026-04-29
graphics/d3d12: add EDRAM resolve diagnostic instrumentation

xdzleo:
# Add EDRAM resolve diagnostic instrumentation

## Summary

This PR adds non-invasive, gated diagnostic instrumentation to `D3D12CommandProcessor::IssueCopy()` so that consumers of rexglue-sdk can observe the EDRAM-resolve-to-CPU-memory pipeline at runtime without forking the SDK.

## Problem statement

When debugging recompiled titles whose swap chain stays black despite high GPU draw activity, there was no way to determine inside the rexglue-sdk's `IssueCopy()` path:

1. **Whether `IssueCopy()` is being entered at all.** The dispatch is gated by `RB_MODECONTROL.edram_mode == kCopy(6)` at `IssueDraw()`. If the title never sets that mode, `IssueCopy()` never fires — but with no logging, this is indistinguishable from "fired but no-op".
2. **What destination address the title is resolving to.** `RB_COPY_DEST_BASE` should match the swap-chain's `fetch_base` for swap-chain resolves; mismatches indicate a different render-target intent.
3. **Whether the readback path (`d3d12_readback_resolve = "fast"` / `"true"`) succeeded.** The pre-existing log line at `command_processor.cpp:3157` only fires for `kDisabled` mode, so users running with the default `kFast` mode get zero feedback.

**Real-world impact (Skate 3 case study):**

When porting Skate 3 (Xbox 360, EA Black Box) to PC via rexglue, the title issued 1,468,834 draws and 13,370 swaps in a 35-second run, but the swap chain at `fetch_base=0x05258000` stayed all-black (`first32B=[00 00 00 FF ...]`). With no observability into `IssueCopy()`, we could not determine whether:

- The title never entered `kCopy` mode (real bug in title-side recompile)
- `kCopy` fired but to a different destination than the swap fetch
- `kCopy` fired correctly but the readback path failed silently

After adding this instrumentation, we determined:
- `IssueCopy()` fires 22,800+ times per run
- 5,351 of those resolves target the swap chain `0x05258000`
- All 22,800 return `resolve_ok=true` with valid `written_address` and `written_length`

This narrowed the root cause to the title-side draw pipeline (the title produces empty draws), not the SDK resolve path. Without this instrumentation, days of investigation would have been wasted on the wrong layer.

## Changes

**File:** `src/graphics/d3d12/command_processor.cpp`
**Lines added:** ~30

The patch wraps `D3D12CommandProcessor::IssueCopy()` with three optional log points:

1. **Entry log:** snapshot of `RB_MODECONTROL`, `RB_COPY_CONTROL`, `RB_COPY_DEST_BASE`, `RB_COPY_DEST_PITCH`, `RB_COPY_DEST_INFO` registers.
2. **kDisabled-path log:** existing log line preserved unchanged (backwards compat).
3. **Readback-path log:** new line for the `kFast`/`kSome` paths showing return value and resolve mode.

All three are gated `if (my_n <= 32 || (my_n % 240) == 0)` for low overhead (~16 lines/min at 60Hz).

### Default behavior

- **No new cvar.** Logging emits at `INFO` level via the existing `REXGPU_INFO` macro.
- **No performance regression.** Counter is `std::atomic<uint64_t>` (one fetch_add per call); register reads are existing operations the function does anyway.
- **No new dependencies.** Uses only headers already included in `command_processor.cpp`.

## Test evidence

Tested in the Skate 3 PC port build. Log output sample (with default gating):

```
[gpu] [IssueCopy] entry #1 MODECONTROL=00000006 COPY_CONTROL=00100000
                       COPY_DEST_BASE=05258000 COPY_DEST_PITCH=02D00500 COPY_DEST_INFO=01000300
[gpu] [IssueCopy] readback-path #1 returned=true mode=1
[gpu] [IssueCopy] entry #240 MODECONTROL=00000006 COPY_CONTROL=00100340
                       COPY_DEST_BASE=04EB1000 COPY_DEST_PITCH=02800480 COPY_DEST_INFO=01000302
[gpu] [IssueCopy] readback-path #240 returned=true mode=1
[gpu] [ReadbackPath] #240 resolve_ok=true written_addr=04EB1000 written_len=294912
...
```

Verified across 3 separate 30-second runs:
- No new crash modes
- No frame-time regression (within 0.5 ms of pre-patch)
- Log file size growth: ~3 KB per minute additional

## Compatibility

- ✅ Works with `d3d12_readback_resolve = "true"` / `"fast"` / `"some"` / `"disabled"`
- ✅ Compiles on MSVC, Clang/LLVM, GCC
- ✅ No header changes
- ✅ No new public API
- ✅ Build-config compatible (Debug, Release, RelWithDebInfo)

## Reviewer checklist

- [ ] Code review of changes in `src/graphics/d3d12/command_processor.cpp`
- [ ] Verify atomics are correctly synchronized
- [ ] Verify gating constants (32, 240) are reasonable for typical 60Hz workloads
- [ ] Consider if log channel should be `REXGPU_DEBUG` instead of `REXGPU_INFO`
- [ ] Optional: add a cvar `d3d12_log_resolves` to enable/disable at runtime

## Related work

This PR is part of a broader effort to improve rexglue-sdk's debuggability for game-porting work. See the [Skate 3 port research repo](https://github.com/xdzleo/skate3-rexglue) for the complete context.

## Future enhancements (not in this PR)

- Add corresponding instrumentation for `IssueCopy_DepthPath()` if it exists
- Consider a `D3D12_RESOURCE_BARRIER` visualization mode for tracking transitions during resolve
- Optional: capture the first 32 bytes of the resolved data to detect "all-zero resolve" cases earlier


tomcl7:
Please review the [Contribution Guidelines](https://github.com/rexglue/rexglue-sdk/wiki/Contributing) before attempting to PR.

All PRs require an issue with proper scoping before we will merge them.

PR #311, opened 2026-04-29
input/mnk: emit keystrokes so XamInputGetKeystroke returns input

### LINKED ISSUE #310: [Bug]: MnK driver: XamInputGetKeystroke always returns X_ERROR_EMPTY
### Description

`MnkInputDriver` declares an `EnqueueKeystroke` helper and a `keystroke_queue_`
but no event handler ever pushes into the queue. As a result, every guest call
to `XamInputGetKeystroke` / `XamInputGetKeystrokeEx` returns `X_ERROR_EMPTY`
whenever MnK is the active input source.

State polling (`XamInputGetState`) is unaffected, so the bug is silent on most
gameplay paths. It only surfaces on UI that uses edge-triggered input — most
visibly, "Press Start" prompts on title screens and 360 dashboard-style menus

### Steps to Reproduce

1. Build a recomp that imports `__imp__XamInputGetKeystrokeEx`.
   (TiP-Recomp does — visible at `generated/retip_init.h:29657`.)
2. Set `mnk_mode = true`, no controller plugged in.
3. Skip past intros with Space (works — that path uses state polling).
4. Reach the "Press Start" title screen and press Escape (the default
   `keybind_start` binding, at least in TiP-Recomp).
5. Nothing happens. Title screen never advances and keyboard inputs aren't detected.

### Expected Behavior

Pressing a key bound to a gamepad button generates a corresponding
`X_INPUT_KEYSTROKE` so guest code that polls keystrokes for "Press Start" /
menu navigation can advance.

### Actual Behavior

Every guest call to `XamInputGetKeystroke` / `XamInputGetKeystrokeEx` returns `X_ERROR_EMPTY`
whenever MnK is the active input source.

### Platform

Windows

### Clang version

_No response_

### Version / Commit

_No response_

_No response_

### Related

_No response_

Zorkats:
Closes #310 

## Input

- Emit keystrokes from `MnkInputDriver` so `XamInputGetKeystroke` / `XamInputGetKeystrokeEx` return input from keyboard and mouse. `GetState` now diffs the freshly computed buttons, triggers, and left-stick direction against tracked previous-frame state and enqueues the matching `VirtualKey::kXInputPad*` code per transition. Right stick (mouse-driven and continuous) is intentionally skipped to avoid flooding the queue; Guide button is omitted because it has no `VK_PAD` code in the standard XInput keystroke spec.

## Fixes

- Populate the `MnkInputDriver` keystroke queue. `EnqueueKeystroke` was defined but never called from any event handler, leaving `keystroke_queue_` permanently empty and causing `XamInputGetKeystroke` / `XamInputGetKeystrokeEx` to always return `X_ERROR_EMPTY` whenever MnK was the input source. The bug was silent on most gameplay (state polling via `XamInputGetState` worked correctly, so WASD movement, intro skipping, and in-game button input all functioned) but broke any title screen or menu that consumed edge-triggered input — "Press Start" prompts appeared unresponsive even though intros could be skipped fine. `GetState` now invokes a new `EmitKeystrokes(buttons, lt, rt, lx, ly)` after computing the gamepad state, inside the existing `state_mutex_` lock. The method table-drives the button bit → `kXInputPad*` mapping (D-pad, Start, Back, A/B/X/Y, both shoulders, both thumb-presses), thresholds triggers digitally (any nonzero deflection emits `kXInputPadLTrigger` / `kXInputPadRTrigger`, sufficient because MnK only ever produces 0 or 0xFF), and classifies the left stick into a 9-state direction enum where diagonals win over cardinals when both axes deflect. `OnLostFocus` resets the new prev-state members alongside the existing `key_down_` reset so resumed focus does not replay stale keystrokes. Tested against TiP-Recomp on 0.7.4 and forward-applied to 0.7.6 (mnk driver files byte-identical between those versions).


tomcl7 (changes requested):
I did confirm that `EnqueueKeystroke` had no callers, but this fix only covers part of what the issue describes though.

- No repeat events. `EmitKeystrokes` only emits on rising and falling edges, but native `XInputGetKeystroke` fires `XINPUT_KEYSTROKE_REPEAT` after an initial delay and at a repeat rate. Issue #310 calls out "360 dashboard style menus", which need repeats for hold to scroll. After this PR a "Press Start" prompt advances, but a hold to scroll menu ticks once per press from MnK and scrolls continuously from a real pad.
- Edge detection at `GetState` time drops fast taps. A press and release that begins and ends between two `GetState` calls leaves the buttons field at 0 in both samples and emits nothing. The transitions are visible in `OnKeyDown` and `OnKeyUp`, which is where edge detection wants to live.
- `EmitKeystrokes` is only called from `GetState`. A title that polls only `XamInputGetKeystroke`, or polls keystrokes much more often than state, won't see anything from MnK.
- `OnLostFocus` resets `prev_*` but doesn't drain `keystroke_queue_`. Before this PR the queue was always empty so this was moot. Now anything queued just before focus loss sits there and gets delivered after focus returns.
- The right stick comment frames the alternative as impossible ("would flood the queue"). It's a tradeoff. Worth saying so, since this silently breaks any title that uses right stick keystrokes for menu navigation.

Zorkats:
I have added your feedback to the fork. Left it commented with clear explanations in case you have any doubts. Tested it and Built it, still works with TiP-Recomp.

tomcl7 (changes requested):
The bug at #310 is real, but the PR body describes a different fix than what's in the diff. It says `GetState` "diffs the freshly computed buttons, triggers, and left-stick direction against tracked previous-frame state" and invokes "a new `EmitKeystrokes(buttons, lt, rt, lx, ly)`." There is no such function in the diff. `KEYDOWN`/`KEYUP` edges actually come from `OnKeyDown` / `OnKeyUp` / `OnMouseDown` / `OnMouseUp` calling `EmitButtonChange` directly https://github.com/rexglue/rexglue-sdk/blob/fe51f7f72a3fe81ac13c113985acb76da62d138a/src/input/mnk/mnk_input_driver.cpp#L489-L504 and `GetState` only ticks repeats and the opt-in right stick https://github.com/rexglue/rexglue-sdk/blob/fe51f7f72a3fe81ac13c113985acb76da62d138a/src/input/mnk/mnk_input_driver.cpp#L223-L232 The trigger description is wrong too: there's no threshold anywhere, triggers route through the same keybind table as A/B/X/Y inside `EmitButtonChange` https://github.com/rexglue/rexglue-sdk/blob/fe51f7f72a3fe81ac13c113985acb76da62d138a/src/input/mnk/mnk_input_driver.cpp#L325-L350

Rewrite the PR description to match what the code actually does and I'll take another look.

tomcl7:
Stale. Reopen when conflicts and requested changes are resolved.

Issue #317, opened 2026-05-01
[Feature]: Wayland support


birabittoh:
## Summary

Most modern Linux desktops run on Wayland. The Rexglue SDK currently requires X11, so games need XWayland to run.

Adding native Wayland support would let games run without any translation layer on the majority of current Linux systems.

## Proposed Solution

When the user is on a Wayland session, use the native Wayland backend instead of falling back to XWayland. The X11 path should remain fully functional for users who are still on X11.

The change should be transparent: no environment variables or manual configuration should be needed for games to work correctly on either session type.

## Requirements

- [x] Detect the active session type at startup and pick the right backend automatically
- [x] Vulkan surface works natively on Wayland (no XWayland)
- [x] Mouse capture and cursor hiding work as expected
- [x] Existing X11 behaviour is unchanged

## Modified Dependencies

- **wayland-client** — new runtime dependency on Linux
- **wayland-protocols** — new build-time dependency on Linux

## Acceptance Criteria

- [x] Game window opens natively on a Wayland desktop without any workarounds
- [x] Mouse capture and look feel identical to the X11 experience
- [x] Forcing X11 via `GDK_BACKEND=x11` still works for users who need it
- [x] SDK builds cleanly on systems without Wayland libraries installed
- [x] Performance on x11 and wayland is not visibly impacted by this change

## Related

- Wayland protocols upstream: https://gitlab.freedesktop.org/wayland/wayland-protocols

monzonit3:
doesn't sdl already support wayland?

PR #319, opened 2026-05-01
feat(wayland): add native Wayland support

birabittoh:
This PR introduces native Wayland support with the following updates and fixes:
- Add native Wayland support via GTK3 with WaylandWindowSurface and a dedicated wl_subsurface for Vulkan rendering.
- Promote WarpPointer to the Window base class to remove platform-specific coupling in the input layer.
- Add error logging for empty input regions and guard XCB fallback to prevent returning invalid XIDs on Wayland sessions.
- Update CI workflow to support Wayland.

Addresses #317.

Tested on Linux (wayland and x11) and Windows 11.


tomcl7 (changes requested):
This is way too many changes wrapped up into a single commit.

I would also like the review of @mrcmunir

Issue #318, opened 2026-05-01
[Bug]: config is serialized with locale decimal separator, breaking TOML config on non-English systems


birabittoh:
### Description

On systems with a locale that uses `,` as the decimal separator (e.g. `it_IT`, `de_DE`, `fr_FR`), double-type cvars are written to the `.toml` config file using `,` instead of `.`.

TOML requires `.` as the decimal separator. A config file written with `,` fails to parse on the next launch, resetting the affected cvar to its default.

### Steps to Reproduce

1. Set system locale to Italian: `export LC_ALL=it_IT.UTF-8`
2. Run the application, press F4, set `mnk_sensitivity` to 1.5, save the config.
4. Inspect the written `.toml`: the value appears as `1,5`.
5. Relaunch the game.

### Expected Behavior

The `.toml` config is always written with `.` as the decimal separator, regardless of the system locale.

### Actual Behavior

The config is written as:

```toml
mnk_sensitivity = 1,5
```

On relaunch:

```
[error] [core] Failed to parse config [...].toml: Error while parsing key-value pair: expected a comment or whitespace, saw ','
```

The TOML config file is completely ignored by the game.

### Platform

Linux

### Clang version

22.1.3

### Version / Commit

fb27737

tomcl7:
Fixed with #332

PR #320, opened 2026-05-01
fix(cvar): use locale-independent serialization for double cvars

birabittoh:
std::to_string(double) respects the C locale but can produce comma decimal separators on systems where LC_ALL/LC_NUMERIC is set to a locale like it_IT or de_DE. TOML requires a period, so the written config file would fail to parse on the next launch.

Replace with a std::to_chars-based helper (DoubleToString) which is locale-independent per the C++ standard.

Addresses #318.

### DIFF
--- a/include/rex/cvar.h
+++ b/include/rex/cvar.h
@@ -235,6 +235,13 @@ inline bool ParseDouble(std::string_view s, double& out) {
   return end != str.c_str() && *end == '\0';
 }
 
+// Locale-independent double → string (TOML requires '.' as decimal separator).
+inline std::string DoubleToString(double v) {
+  char buf[32];
+  auto [ptr, ec] = std::to_chars(buf, buf + sizeof(buf), v);
+  return ec == std::errc{} ? std::string(buf, ptr) : std::to_string(v);
+}
+
 }  // namespace rex::cvar
 
 //=============================================================================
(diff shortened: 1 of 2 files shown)


tomcl7:
This is fixing the symptom, not the cause, and only at two of the call sites that have the problem. The reason `std::to_string` writes a comma is that `gtk_init_check` calls `setlocale(LC_ALL, "")` during init, flipping the entire process's C locale to whatever `LC_ALL`/`LANG` resolve to. That affects every locale sensitive C function for the rest of the run, including `strtod`, `atof`, `printf("%f", ...)`, `sscanf`, and anything in third-party code that goes through them. A few other sites in this repo are exposed to the same locale by way of `strtod`/`atof`:

https://github.com/rexglue/rexglue-sdk/blob/2250ee041cf3068aace5b95539ab84f2037369bb/src/ui/overlay/settings_overlay.cpp#L472
https://github.com/rexglue/rexglue-sdk/blob/2250ee041cf3068aace5b95539ab84f2037369bb/include/rex/cvar.h#L234
https://github.com/rexglue/rexglue-sdk/blob/2250ee041cf3068aace5b95539ab84f2037369bb/include/rex/string/util.h#L228

`ApplyTomlTable` in `src/core/cvar.cpp` also runs `std::to_string` on TOML doubles before feeding them to `ParseDouble`'s `strtod`, which round trips correctly today only because both functions are locale-sensitive in the same direction. That's the kind of accident that breaks silently the day someone swaps `ParseDouble` for `std::from_chars`. Same locale, same bug, none of it visible from this diff. Taking this PR pins us into a pattern where every new double cvar and every new `strtod` site in the SDK turns into a fresh round of whack-a-mole.

GTK provides a documented opt-out for exactly this case. From `gtk_disable_setlocale` (https://docs.gtk.org/gtk3/func.disable_setlocale.html):

> Prevents `gtk_init()`, `gtk_init_check()`, `gtk_init_with_args()` and `gtk_parse_args()` from automatically calling `setlocale (LC_ALL, "")`.
>
> You would want to use this function if you wanted to set the locale for your program to something other than the user's locale, or if you wanted to set different values for different locale categories.

The corresponding gate in the GTK 3 source lives in `setlocale_initialization()` in `gtk/gtkmain.c`, controlled by a `do_setlocale` flag that this function clears.

So the fix is one call before `gtk_init_check` in `src/ui/windowed_app_main_posix.cpp`:

```cpp
gtk_disable_setlocale();
if (!gtk_init_check(&argc_post_gtk, &argv_post_gtk)) {
  ...
}
```

That keeps the process in the default `C` locale, which is what every cvar getter, every `strtod`, and `std::to_string` in the SDK already assume. The cvar code stays as is, and #318 closes for the right reason. If down the line we want localized strings or dates from glib/GTK, the doc-recommended path is to keep `gtk_disable_setlocale()` and call `std::setlocale` ourselves on the specific categories we want, leaving `LC_NUMERIC` alone.

Going to close this PR. Would you mind opening a new one against `development` with the one line `gtk_disable_setlocale()` change and linking #318? Once that's merged I'll close the issue out.

PR #330, opened 2026-05-06
Emit type / enum headers from config-supplied data

### LINKED ISSUE #329: [Feature]: Emit type / enum headers from config-supplied [[types]] / [[enums]]
### Summary

Recompiled output today is wired up via the `[functions]` table — addresses, sizes, and human names. The host substrate that wraps that output (kernel / audio / graphics shims) still has to address PPC-side data with hand-coded byte offsets:

```cpp
// kernel shim: clear the alerted byte on a thread
*((uint8_t*)thread + 0x6D) = 0;
```

The same producer that generates the `[functions]` data (a PDB extractor, in our case) carries struct / union / class layout and enum definitions. Surfacing those alongside function names lets the same shim become:

```cpp
*((uint8_t*)thread + _KTHREAD_offsets::Alerted) = 0;
```

— self-documenting, drift-detectable (the offset constant fails the build if the producer ever changes), zero-overhead at runtime.

This is the natural follow-up to the existing `[functions]` override mechanism: same producer, same idiom, separate top-level array.

### Proposed Solution

Extend the existing config schema with two top-level arrays-of-tables, `[[types]]` and `[[enums]]`, parsed exactly like `[[switch_tables]]` / `[[invalid_instructions]]`. A new emit step writes:

```
<out_directory_path>/mappings_generated/types.h
<out_directory_path>/mappings_generated/enums.h
```

alongside the recompiler's `.cpp` files. Pure declarations — `<cstddef>` / `<cstdint>` only.

#### Key features
- **Opaque-storage types**: each entry emits `struct <Name> { unsigned char _opaque[size]; };` plus `static_assert(sizeof) == size` and a sibling `<Name>_offsets` namespace of `inline constexpr size_t` field offsets.
- **`enum class` enums** with fixed-width underlying types. `bool` / `int8_t` / `uint8_t` / .../ `uint64_t` mapping for producer strings.
- **Sanitization**: `::` → `__`, non-id bytes → `_`, anonymous tags (`<unnamed-tag>`) skipped, dedupe by sanitized identifier first-wins, C++ keyword guard on field names.
- **No new gate / CLI flag**: emit when data is present, no-op when absent. Matches the `[functions]` semantics.
- **Layered config support**: `[[types]]` / `[[enums]]` can live in the main config or any file pulled in via `includes`.

#### Schema
- [ ] `RecompilerType` / `RecompilerTypeField` / `RecompilerEnum` / `RecompilerEnumValue` structs in `include/rex/codegen/config.h`.
- [ ] `types` / `enums` vectors on `RecompilerConfig`.
- [ ] Sanitizer helper in anonymous namespace alongside the existing function-name sanitizer.

#### Parsing
- [ ] `[[types]]` and `[[enums]]` array parsing in `ApplyToml` (`src/codegen/config.cpp`), mirroring existing `[[switch_tables]]` patterns.
- [ ] First-wins dedupe on sanitized identifier, with skip / collision counts logged.
- [ ] Post-load summary log lines.

#### Emission
- [ ] New `EmitTypeHeaders(const CodegenContext&)` free function (header + cpp).
- [ ] Helpers: enum-underlying mapping table, comment escape (`*/` → `* /`), enum-value sanitizer, C++ keyword guard, field-name sanitizer.
- [ ] Output written under `<configDir>/<outDirectoryPath>/mappings_generated/`.
- [ ] Pipeline integration: invoke between `Analyze` and `CodegenWriter::write` in `CodegenPipeline::Run`.

#### Documentation
- [ ] `docs/types_and_enums.md` covering schema, sanitization, layout, underlying-type mapping, and an end-to-end example.

### Out of Scope

- **Named-field emission.** Translating producer type expressions (e.g. `void(_KTHREAD*, unsigned long)*`, `_PROCESS_HEAP_ENTRY::<unnamed-tag>::<unnamed-type-Block>`) into self-consistent C++ requires dialect-aware parsing plus topological ordering of UDTs. Several months of work; tracked separately. Until landed, host code uses the `<Type>_offsets` namespace.
- **Static-assert per field offset.** Possible to emit `static_assert(<Type>_offsets::Foo == 0xN)` for each field, but redundant with the `inline constexpr` value itself. Considered and dropped.
- **Producer-format-specific schemas.** The proposed schema fields (`name`, `size`, `kind`, `unique_name`, `[[types.fields]]`) describe layout, not provenance. Producers can omit any optional field they don't carry.

### Modified Dependencies

None. Uses existing `toml++` for parsing and `fmt` for emission.

### Acceptance Criteria

- [ ] No-config behaviour unchanged: codegen with no `[[types]]` / `[[enums]]` produces zero `mappings_generated/` output.
- [ ] Hand-crafted smoke fixture covering normal types, anonymous tags, sanitized collisions, forward-declared types, scoped names, keyword fields, comment-escape type exprs, anonymous-union flattening, multiple enum underlyings (`int`, `unsigned char`, `bool`), and empty enums emits cleanly.
- [ ] Generated headers compile under `clang++ -std=c++23 -c -o /dev/null` against a TU that `#include`s both.
- [ ] `docs/types_and_enums.md` describes the schema, sanitization rules, layout, underlying mapping, and an end-to-end example.

### Related

- Existing `[functions]` mechanism (the precedent this extends): `src/codegen/config.cpp` ApplyToml + `src/codegen/phase_register.cpp` config-functions block.
- Existing array-of-tables idioms to mirror: `[[switch_tables]]`, `[[invalid_instructions]]`, `[[midasm_hook]]`.
- PR with the implementation will reference this issue.

ItsNotPaths:
Closes #329.

> **Stacked on top of #327** (`--force` analyze fix). The diff against `main` includes the `--force` commit until #327 merges; once it does, GitHub will auto-rebase this PR and the diff will collapse to just the type/enum work.

## Summary

Add an emission step that materialises config-supplied `[[types]]` and `[[enums]]` arrays into two header files alongside the recompiled `.cpp` output:

```
<out_directory_path>/mappings_generated/types.h
<out_directory_path>/mappings_generated/enums.h
```

Each producer type becomes an opaque ` struct` with the right size, a sibling `<Name>_offsets` namespace of `inline constexpr size_t` field offsets, and a `static_assert(sizeof) == size` drift check. Each producer enum becomes an `enum class` with a fixed-width underlying type. Pure declarations — `<cstddef>` / `<cstdint>` only.

The motivation is the same one that drives `[functions]` overrides: a producer (typically a PDB extractor) carries layout / enum data alongside the function names, and we want the host substrate to address PPC-side data via named offset constants instead of hand-coded byte offsets:

```cpp
// before
*((uint8_t*)thread + 0x6D) = 0;

// after
*((uint8_t*)thread + _KTHREAD_offsets::Alerted) = 0;
```

Self-documenting, drift-detectable (the constant fails the build if the producer ever changes), zero-overhead at runtime.

## Schema

`[[types]]` and `[[enums]]` are top-level arrays-of-tables, parsed from any config file (main or included via `includes`), with the same merge semantics as `[[switch_tables]]` / `[[invalid_instructions]]`. Full schema reference is in `docs/types_and_enums.md` (added in this PR).

```toml
[[types]]
name        = "Vehicle"
unique_name = ".?AVVehicle@@"
kind        = "class"
size        = 0x80

[[types.fields]]
name      = "rpm"
type_expr = "float"
offset    = 0x10

[[enums]]
name        = "VehicleClass"
underlying  = "int"

[[enums.values]]
name  = "Sedan"
value = 0
```

## Design decisions

- **No new gate / CLI flag.** Emit when data is present, no-op when absent. Matches `[functions]` semantics — data drives behaviour, not a separate switch.
- **Schema lives in the existing config**, not a new `*_file_path` key. Use `includes` for layered configs.
- **Plural array names** (`[[types]]`, `[[enums]]`, `[[types.fields]]`, `[[enums.values]]`) to match rexglue's existing `[[switch_tables]]` / `[[invalid_instructions]]` idiom.
- **Opaque storage, not named-field translation.** Translating producer type expressions (`unsigned long:2`, `void(_KTHREAD*, unsigned long)*`) into self-consistent C++ requires dialect-aware parsing plus topological ordering of UDTs. Substantial; intentionally deferred. Field offsets cover the host substrate's actual need.
- **Sanitization with no address suffix.** Types and enums are name-keyed (unlike functions, which can share `display` names across overloads). Anonymous tags (`<unnamed-tag>`) are skipped. Sanitized-identifier collisions drop the second occurrence first-wins.

## Files

| File | Change |
|---|---|
| `include/rex/codegen/config.h` | Add `RecompilerType` / `RecompilerTypeField` / `RecompilerEnum` / `RecompilerEnumValue` structs and `types` / `enums` vectors on `RecompilerConfig`. |
| `src/codegen/config.cpp` | `SanitizeTypeName` helper; `[[types]]` / `[[enums]]` parsing in `ApplyToml`; post-load summary log lines. |
| `include/rex/codegen/emit_type_headers.h` | Public entry point declaration. |
| `src/codegen/emit_type_headers.cpp` | New file. `EmitTypeHeaders` body plus helpers (`EnumUnderlying`, `EscapeComment`, `SanitizeEnumValue`, `IsCppKeyword`, `SanitizeFieldName`). |
| `src/codegen/codegen.cpp` | Invoke `EmitTypeHeaders(*ctx_)` between Analyze and `CodegenWriter::write`. |
| `src/codegen/CMakeLists.txt` | Add `emit_type_headers.cpp` to the rexcodegen sources. |
| `docs/types_and_enums.md` | New user-facing schema reference. |

### Smoke test (hand-crafted fixture)

A small TOML fixture covers normal struct with fields, anonymous-tag skip, sanitized identifier collision drop, forward-declared (`size = 0`) type, scoped name (`::` → `__`), keyword guard on a field name (`register` → `f_register`), comment-escape-friendly `type_expr`, anonymous-union flattening (two fields at same offset, different names — both surface), sanitized enum value with weird chars, duplicate enum value drop, multiple integer underlyings (`int`, `unsigned char`, `bool`), and an empty enum.

```
[info] [codegen] Loaded 4 type definitions for header emission
[info] [codegen] Loaded 5 enum definitions for header emission
[info] [codegen] [type-headers] wrote out/mappings_generated/types.h (4 types)
[info] [codegen] [type-headers] wrote out/mappings_generated/enums.h (5 enums)
```

(One type collision-dropped, one anonymous-tag-skipped, three emitted; five enums all emitted.)

Generated headers compile clean:

```
$ clang++ -std=c++23 -c -o /dev/null types_enums_check.cpp -I out/mappings_generated/
$ echo $?
0
```

### Real-data validation

Pending downstream: a PDB extractor on our side is being extended to emit `[[types]]` / `[[enums]]` blocks on top of its existing `[functions]` output. Once that lands we'll have a real-data run to drop in here (~50k types, ~4k enums against the FH1 retail XEX) before merge.

## Compatibility

- Default behaviour unchanged when neither `[[types]]` nor `[[enums]]` is configured. No `mappings_generated/` directory is created.
- The sanitized-identifier scheme is stable: same input config produces the same headers byte-for-byte. Re-runs are idempotent.
- Existing config files are unaffected; older builds silently ignore the new top-level arrays.

## Out of scope

- **Named-field emission.** Future opt-in mode that parses producer type expressions, topologically sorts UDTs, and emits real `struct Foo { _BAR baz; };` definitions. Tracked separately.
- **Static-assert per field offset.** Possible but redundant with the `inline constexpr` value itself.

### DIFF
--- a/docs/types_and_enums.md
+++ b/docs/types_and_enums.md
@@ -1,0 +1,256 @@
+# Type / enum header emission
+
+Optional codegen step that materialises producer-supplied
+`[[types]]` / `[[enums]]` config arrays into two header files
+alongside the recompiler's `.cpp` output:
+
+```
+<out_directory_path>/mappings_generated/types.h
+<out_directory_path>/mappings_generated/enums.h
+```
+
+The headers are pure declarations — they pull in `<cstddef>` /
+`<cstdint>` and nothing else. Each producer type becomes an opaque
+`struct` of the right size, guarded by `static_assert(sizeof(...) ==
+size)`, with a sibling `<Type>_offsets` namespace whose members are
+`inline constexpr size_t` field offsets. Each producer enum becomes
+a strongly-typed `enum class` with a fixed-width underlying type.
+
+## When to use this
+
+A mapping file is the right tool when the host substrate (audio /
+graphics / kernel HAL code) has to address PPC-side data with
+hand-coded byte offsets:
+
+```cpp
+// kernel shim: clear the alerted byte on a thread
+*((uint8_t*)thread + 0x6D) = 0;
+```
+
+With type headers emitted from the same producer the recompiler
+already trusts for function names, the same shim becomes:
+
+```cpp
+*((uint8_t*)thread + _KTHREAD_offsets::Alerted) = 0;
+```
+
+Self-documenting, drift-detectable (the constant fails the build if
+the producer ever changes), zero-overhead at runtime.
+
+This feature is *not* required for codegen and is a no-op unless the
+config supplies `[[types]]` and/or `[[enums]]` arrays.
+
+## Schema
+
+`[[types]]` and `[[enums]]` are top-level arrays-of-tables, parsed
+from any included config file with the same merge semantics as
+`[[switch_tables]]` and `[[invalid_instructions]]`. Place them in
+the project's main config or in a separate file pulled in via
+`includes`.
+
+### `[[types]]`
+
+| Field | Type | Required | Use |
+|-------|------|----------|-----|
+| `name` | string | yes | Display name (`_KTHREAD`, `Vehicle::State`). Sanitized into a C++ identifier. Anonymous tags (`<unnamed-tag>`) are skipped. |
+| `size` | uint32 | yes (effectively) | sizeof, in bytes. `0` skips the static_assert and uses `_opaque[0x1]`. |
+| `kind` | string | no | `"struct"` / `"union"` / `"class"`. Header comment only; emitted as `struct` regardless (opaque storage). Defaults to `"struct"`. |
+| `unique_name` | string | no | Producer's mangled / unique name. Header comment only. |
+| `[[types.fields]]` | array | no | Per-field offset entries. |
+
+Per `[[types.fields]]` entry:
+
+| Field | Type | Required | Use |
+|-------|------|----------|-----|
+| `name` | string | yes | Sanitized to a C++ identifier; reserved-word names (`register`, `not`, ...) are prefixed with `f_`. |
+| `offset` | uint32 | yes | Bytes from the start of the parent type. |
+| `type_expr` | string | no | Raw producer type expression. Preserved verbatim and emitted as a comment beside the offset constant. *Not parsed* — translating producer type expressions into self-consistent C++ is intentionally out of scope. |
+
+### `[[enums]]`
+
+| Field | Type | Required | Use |
+|-------|------|----------|-----|
+| `name` | string | yes | Display name. Sanitized; anonymous tags skipped. |
+| `underlying` | string | no | Producer's integer type (`int`, `unsigned long`, `char`, `unsigned char`). Defaults to `int`. Translated to a fixed-width type at emit time (see "Underlying-type mapping"). |
+| `unique_name` | string | no | Header comment only. |
+| `[[enums.values]]` | array | yes (non-empty for non-empty enums) | Per-enumerator entries. |
+
+Per `[[enums.values]]` entry:
+
+| Field | Type | Required | Use |
+|-------|------|----------|-----|
+| `name` | string | yes | Sanitized to a C++ identifier. |
+| `value` | int64 | yes | Signed-widened — fits both signed and unsigned underlyings. Narrowing happens at the `enum class` boundary. |
+
+## Sanitization
+
+Type / enum / nested-name sanitizer (separate from the function-name
+sanitizer):
+
+1. Anonymous producer names (anything starting with `<`, e.g.
+   `<unnamed-tag>`) are skipped — they cannot be surfaced at file
+   scope without a synthesized stable identifier, and the producer
+   doesn't give us one.
+2. `::` collapses to `__` (preserves the scope marker).
+3. Each remaining non-`[A-Za-z0-9_]` byte becomes `_`.
+4. Empty / leading-digit results are dropped (counts as
+   "anonymous or unsanitizable").
+
+There is intentionally **no address suffix** — types and enums are
+name-keyed, not address-keyed. When two distinct producer entries
+sanitize to the same identifier, the first wins; subsequent
+collisions are dropped silently (logged at `[debug]`).
+
+Field names use a stricter sanitizer (no `::` handling) plus a C++
+keyword guard: a sanitized field name that collides with a reserved
+word gets an `f_` prefix. Enum value names skip the guard because
+producer enumerators are conventionally upper-case and collisions
+don't arise in practice.
+
+## Layout rules
+
+### Types
+
+```cpp
+// <raw_name> (kind=<kind>, unique_name=<unique>)
+struct <sanitized_name> {
+    unsigned char _opaque[0x<size>];
+};
+static_assert(sizeof(<sanitized_name>) == 0x<size>, "<name> size mismatch with producer");
+namespace <sanitized_name>_offsets {
+    inline constexpr size_t <FieldA> = 0x<off>; // <raw_field_type>
+    // ...
+}
+```
+
+The opaque-buffer choice is what makes this a tractable feature
+rather than a multi-month type-system project. Consumers that need
+named-field access cast through the offset constants:
+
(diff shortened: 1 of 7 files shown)


ItsNotPaths:
tested against full mapping with mapped types and enums, correct output through and through

tomcl7:
The motivating example here, `_KTHREAD_offsets::Alerted`, is already covered by `X_KTHREAD` in `include/rex/system/xthread.h`, which has `alerted[2]` at offset 0x6D and a `static_assert_size(X_KTHREAD, 0xAB0)` already doing drift detection. The realistic version of the "before" snippet in the PR description is `thread->alerted[0] = 0`, not a raw byte pointer plus a hex literal.

- I don't follow the use case. Every motivating example you've shown is already handled, and handled better, by an SDK struct someone wrote by hand with real field types and a size assert. Generating an opaque blob plus an `_offsets::` namespace gives strictly less than what `X_KTHREAD`, `X_KPRCB`, `X_DISPATCH_HEADER` etc. already provide. You lose the field types and any ability to chase pointers, in exchange for a constant that resolves to the same hex offset someone would have written into the struct definition anyway.
- The before / after comparison in the PR body is misleading once you look at the actual SDK. Nobody writing host substrate code reaches for `*((uint8_t*)thread + 0x6D)` when `thread->alerted[0]` is sitting there. Compared against the realistic baseline, the after is a regression.
- 812 lines of generator plus a 256 line schema document for output that's strictly downstream of what we already write by hand. If a someone is extracting layout data, the right destination is a curated header under `include/rex/` that uses the layout to inform a typed struct, the same way the existing system headers were authored. If it is a project specific structure, the same concept applies but local to your project. 
- The "named field emission is intentionally deferred" disclaimer is doing all the work in the PR description. Named field structs are the only version of this that would actually be useful, and they're the version we already write by hand. Generating the opaque intermediate step doesn't create anything that isn't already available.

Closing this. To put it directly, reviewing 800+ lines of generator code that duplicates what the SDK already does isn't a productive use of review time. Before opening another PR at this size, spend some time actually looking at what can be improved or fixed in the existing code, rather than adding entirely new systems without prior a discussion with myself and the core contributors.

Issue #294, opened 2026-04-19
[Feature]: Dump and replace Textures


Graine25:
### Summary

The feature adds two complementary systems to the ReXGlue SDK's graphics pipeline: a texture dumper that saves decoded Xbox 360 textures to disk as PNG files, and a custom texture loader that intercepts texture loads and substitutes user-provided replacement images. This is needed to enable modding workflows (texture packs, upscaling) and debugging of texture rendering issues in statically recompiled Xbox 360 titles. Without it, there is no way to inspect or override the raw textures being decoded from guest memory by the TextureCache.

### Proposed Solution

The dumper should be a self-contained helper class (similar to the existing `Shader::DumpUcode` pattern in `src/graphics/pipeline/shader/shader.cpp`) that hooks into `TextureCache::CommitPreparedTextureLoad` in `src/graphics/pipeline/texture/cache.cpp` after a texture is successfully uploaded. `cache.cpp:412-446` It should track already-dumped textures in an `std::unordered_set` keyed on a hash derived from TextureKey to avoid redundant writes, and write PNG output using the already-present `stb_image_write` dependency (used in `src/graphics/trace_dump.cpp`). `trace_dump.cpp:26-141` The custom loader should scan a configurable directory at startup, build a `std::unordered_map<TextureKey, std::filesystem::path>`, and be checked inside FindOrCreateTexture before falling through to the normal GPU-side decode path. `cache.cpp:855-876` Both features should be gated by new REXCVAR_DEFINE_BOOL / REXCVAR_DEFINE_STRING flags added to src/graphics/flags.cpp and declared in `include/rex/graphics/flags.h`, following the same pattern as dump_shaders. `flags.cpp:27`

### Requirements

- [ ] New CVARs: `dump_textures` (bool), `dump_textures_path` (string), `load_custom_textures` (bool), `custom_textures_path` (string) added to `src/graphics/flags.cpp` and `include/rex/graphics/flags.h`
- [ ] `TextureDumper` helper class that tracks dumped texture hashes and writes PNG via `stbi_write_png_to_func`
- [ ] Readback path to copy a decoded D3D12 texture resource to a CPU-readable staging buffer for PNG export
- [ ] Integration point in `TextureCache::CommitPreparedTextureLoad` (or `D3D12TextureCache::LoadTextureDataFromResidentMemoryImpl`) to call the dumper after a successful load
[...]

PR #333, opened 2026-05-08
Feature(GPU): Texture dumper and replacer

Graine25:
This feature adds a texture dump and replacement system that identifies guest textures by a stable content hash, dumps them to `textures/dump`, and loads user-authored replacements from `textures/replace`. When enabled through CVars, matching replacements are decoded to RGBA8 and injected into the D3D12 texture cache in place of the original guest texture.

- `build(thirdparty): expose stb_image target` - Adds stb_image as an available third-party target and updates install handling so the replacement pipeline can decode image files, especially PNG replacements, without embedding a separate image dependency path.

 - `a8517da feat(gpu): add texture replacement pipeline` - Introduces the core replacement/dumper module under `graphics/pipeline/texture`. This adds hashing, replacement directory scanning, PNG/DDS loading, DDS dumping, guest texture untiling/endian conversion for dumps, and the filesystem layout: `textures/dump` for generated assets and `textures/replace` for user-provided replacements.

- `7a90eaa feat(gpu): inject replacement textures` - Wires the replacement pipeline into the texture cache and D3D12 upload path. It hashes guest texture contents, looks up matching replacement files, adjusts the texture key for RGBA8 replacement uploads, preserves the original guest layout for memory watches, and keeps temporary upload buffers alive until the GPU has finished using them.

- `feat(gpu): add texture replacement cvars` - This commit adds the user-facing controls for the texture edit feature: texture_dump_enabled, `texture_replace_enabled`, and `texture_folder`. Dumping and replacement are now gated by CVars, and `texture_folder` lets users choose a custom root directory for `dump/` and `replace/` instead of always using `<executable folder>/textures`.

More effort can be made to implement this into Vulkan as well as adjusting for more DDS formats or even PNG's, but this has mostly been optmized for D3D12. Almost all work is @SolarCookies. Addresses issue #294


tomcl7:
I like the idea and it's very cool but I think this adds too much code to maintain given the current size of the team.

Maybe in the future we can look at implementing something similar but I am afraid given the current state this will be a pain to handle as we evolve things.

Thank you for submitting this anyways.

PR #335, opened 2026-05-09
v0.8.0 - Multi-Binary DLL Support

### LINKED ISSUE #264: [Feature]: DLL Support
### Summary

Implement multi-binary and DLL loading support in the codegen pipeline and runtime to enable recompilation of Xbox 360 titles that ship multiple binaries.

The current system assumes a single binary:
- one XEX
- one config
- one output folder compiled directly into the executable

This means we can't:
- Declare a multi-binary project in the config
- Generate recompiled code for DLLs as separate compilation units
- Satisfy cross-module indirect calls (main XEX calling DLL functions obtained via GetProcAddress)
- Call DLL entry points (DllMain) during load/unload

### Proposed Solution

Introduce a **manifest config** and **dynamic shared library loading** for DLL modules.

```
default.xex    -->  projectname_default_xex.lib    (static, linked into game exe)
somelib.dll    -->  projectname_somelib_dll.dll     (shared, loaded at runtime)
anotherlib.dll -->  projectname_anotherlib_dll.dll  (shared, loaded at runtime)
```

The entrypoint module compiles as a **static library** (always loaded, never unloaded). DLL modules compile as **shared libraries** (`.dll`/`.so`) loaded/unloaded at runtime via `LoadLibrary`/`dlopen` to match Xbox 360 semantics. Many titles rely on dynamic module loading, some unload all DLLs and reload different variants when switching game modes or characters. Multiple DLL variants may load at the same guest base address at different times.

### Manifest Config

A new top-level TOML manifest replaces the single-binary config as the codegen entry point:

```toml
# mygame_manifest.toml
[project]
projectName = "mygame"

[entrypoint]
config = "mygame_default_xex.toml"

[[modules]]
config = "mygame_somelib_dll.toml"
guestPath = "bin/somelib.dll"

[[modules]]
config = "mygame_anotherlib_dll.toml"
guestPath = "data/anotherlib.dll"
```

- `project.projectName` - project name for output naming.
- `entrypoint.config` - path (relative to manifest) to the entrypoint per-binary config. Always built, always statically linked.
- `modules[].config` - path to DLL per-binary config.
- `modules[].guestPath` - the path the guest passes to `XexLoadImage`. Used for runtime matching (case-insensitive, backslash-normalized, device-prefix-stripped).

Per-binary configs retain the existing `RecompilerConfig` format unchanged. The manifest just ties them together.

DLL flag is auto-detected from `XEX_MODULE_DLL_MODULE` in the XEX header, with an optional `isDll = true/false` config override.

#### Unloaded Module Trap

All hash table slots are initialized to a trap function. A new unconditional field `uint32_t last_indirect_target` is added to `PPCContext` (unconditional because `ctr` may be optimized to a local variable via `PPC_CONFIG_CTR_AS_LOCAL`). The `PPC_CALL_INDIRECT_FUNC` macro stores the guest target in this field before dispatch:

```cpp
void __unloaded_module_trap(PPCContext& ctx, uint8_t* base) {
  FATAL("Call to unloaded or unregistered function at guest address 0x{:08X}",
        ctx.last_indirect_target);
}
```

### Codegen Pipeline

The pipeline gains a new orchestration layer above the existing per-binary `CodegenContext`:

1. **Load manifest** - parse TOML, resolve per-binary config paths.
2. **Load all binary headers** - `BinaryView` per module extracts code base, code size, image base, image size, PE module name, DLL flag.
3. **Collect all function addresses** - from analysis + pre-assigned thunks.
4. **Compute perfect hash** - emit warning if load factor or slot count exceeds thresholds.
5. **Generate per-binary** - existing codegen pipeline per targeted binary.
6. **Emit shared `ppc_config.h`** - hash constants, thunk region base/max.
7. **Emit per-binary registration function** - replaces `function_table_init.cpp` / `ppc_func_mapping.cpp`.
8. **Emit module registry** - `generated/module_registry.cpp` mapping guest paths to shared libs.
9. **Emit build targets** - static lib for entrypoint, shared libs for DLLs.

Each DLL shared library exports `ReXModule_Register` via platform-appropriate export mechanism:

```cpp
#ifdef _WIN32
#define REX_MODULE_EXPORT __declspec(dllexport)
#else
#define REX_MODULE_EXPORT __attribute__((visibility("default")))
#endif

extern "C" REX_MODULE_EXPORT
void ReXModule_Register(rex::runtime::FunctionDispatcher* dispatcher) {
  dispatcher->SetFunction(0x83000100, &sub_83000100);
  // ...
}
```

Invocation:
```
rexglue codegen manifest.toml                          # build all binaries
rexglue codegen manifest.toml --target somelib_dll     # build default_xex + somelib_dll
```

Entrypoint is always included regardless of `--target`.

### FunctionDispatcher Changes

- `InitializeFunctionTable(uint32_t hash_table_base, uint32_t hash_table_slots)` - allocates hash table in kernel space, initializes all slots to trap function.
- `SetFunction(uint32_t guest_address, PPCFunc* func)` - hashes address, writes to guest memory hash table (single write). C++ `function_table_` map maintained separately for `Execute`/`GetFunction`.
- `RegisterModule(const std::string& module_id, RegisterFn register_func)` - enables recording mode, calls registration function, captures all addresses written via `SetFunction` into `module_addresses_[module_id]`.
- `UnregisterModule(const std::string& module_id)` - overwrites recorded addresses' slots with trap function. Also invalidates thunks allocated for the module's exports.
- `AllocateThunk` - unchanged interface, allocates from pre-assigned kernel space range. Thunks invalidated on unload but not freed; new thunks allocated on reload.
- Thread safety: 8-byte aligned slots, atomic on x86-64. DLL loading is synchronous on loader thread.

### Module Registry and Registration

Generated `module_registry.cpp` in `generated/` root:

```cpp
struct RecompiledModuleInfo {
  std::string pe_name;          // PE module name from XEX header
  std::string guest_path;       // guest path for XexLoadImage matching
  std::string shared_lib_name;  // shared library base name
  RegisterFn register_func;     // non-null for static entrypoint only
};
```

`KernelState::RegisterRecompiledModule` stores entries in `recompiled_modules_`. `RegisterRecompiledModules` call is folded into the existing generated init path so `main.cpp` never changes when modules are added/removed.

### Runtime Module Loading Flow

**Startup:**
1. `RegisterRecompiledModules` registers all module metadata with `KernelState`.
2. `Runtime::Setup` initializes hash table in kernel space.
3. Entrypoint `LoadFromFile` loads XEX data sections (code sections are dead but loaded for header parsing).
4. Entrypoint registration function populates hash table.
5. `Runtime::LaunchModule` starts execution.

**DLL Loading (`XexLoadImage`):**
1. `LoadUserModule` loads XEX data sections into guest memory.
2. Match guest path against registered modules.
   - Match found: `LoadLibrary`/`dlopen` shared lib, resolve `ReXModule_Register`, call it.
   - No match: warn and fail with `X_STATUS_OBJECT_NAME_NOT_FOUND`.
3. If DLL with entry point: call DllMain with `DLL_PROCESS_ATTACH` via `FunctionDispatcher::Execute` - `r3` = hmodule, `r4` = 1, `r5` = 0.
4. Return hmodule.

**DLL Unloading (`XexUnloadImage`):**
1. Call DllMain with `DLL_PROCESS_DETACH` - `r3` = hmodule, `r4` = 0, `r5` = 0.
2. `UnregisterModule` overwrites hash table slots + thunks with trap function.
3. `FreeLibrary`/`dlclose` shared library.
4. Existing unload path: decrement load count, remove from `user_modules_`.

**GetProcAddress (`XexGetProcedureAddress`):**
1. Resolve ordinal to guest address via module export table.
2. Allocate thunk from pre-assigned kernel space range.
3. Write function pointer to thunk's pre-computed hash table slot.
4. Guest calls through thunk address; `PPC_LOOKUP_FUNC` resolves via hash.

### CLI Changes

- `rexglue init` - alias for `rexglue init project`. Scaffolds manifest-based project with manifest TOML, entrypoint config, CMakeLists, presets, main.cpp, app header, `rexglue.cmake`.
- `rexglue init module <binary>` - adds DLL module to existing project: creates per-binary config, registers in manifest under `[[modules]]`, updates `rexglue.cmake`.
- `rexglue codegen` - accepts manifest TOML (detected by `[entrypoint]` key) or legacy single-binary config (backwards compatible).

### Backwards Compatibility

- Single-binary projects work unchanged. Codegen detects manifest vs legacy config.
- Per-binary config format is unchanged. Manifests compose them without modification.
- `rexglue init` without subcommand remains an alias for `init project`.
- **Breaking: generated output format changes.** `function_table_init.cpp` / `ppc_func_mapping.cpp` replaced by registration functions. `ppc_config.h` emits hash constants instead of spanning range constants.
- **Breaking: `PPC_LOOKUP_FUNC` macro changes** from address-indexed to hash-based in `context.h`. Unconditional - no dual-mode guard. All recompiled code must be regenerated after SDK update.
- **Breaking: `PPC_CALL_INDIRECT_FUNC` macro changes** to store target in new `PPCContext::last_indirect_target` field.

### Manifest and Config
- [ ] New manifest TOML format with `[project]`, `[entrypoint]`, `[[modules]]` sections
- [ ] `guestPath` field per module for runtime matching (case-insensitive, backslash-normalized, device-prefix-stripped)
- [ ] Per-binary configs retain existing `RecompilerConfig` format unchanged
- [ ] Config path resolution relative to manifest directory
- [ ] Auto-detect `XEX_MODULE_DLL_MODULE` from XEX header with optional `isDll` override

### Perfect Hash Function Dispatch
- [ ] Codegen collects all function addresses across all declared modules plus pre-assigned thunk addresses
- [ ] Multiply-shift perfect hash construction: 50% initial load factor, expand after 10K candidates, error at 4M slots
- [ ] Hash table allocated at fixed kernel space address `0x80040000` in v80000000 heap
- [ ] `PPC_LOOKUP_FUNC` macro updated to hash-based lookup in `context.h`
- [ ] `PPC_CALL_INDIRECT_FUNC` macro stores target in new `PPCContext::last_indirect_target` field
- [ ] Thunk addresses pre-assigned at `0x80080000` (default 4096 slots), included in hash computation
- [ ] Trap function for all unregistered/unloaded slots with diagnostic output via `last_indirect_target`
- [ ] All hash table slots initialized to trap function at startup

### Codegen Pipeline
- [ ] Orchestration layer above `CodegenContext` for multi-binary manifest builds
- [ ] Shared `ppc_config.h` with hash constants emitted for all modules
- [ ] Per-binary registration function replacing `function_table_init.cpp` / `ppc_func_mapping.cpp`
- [ ] Project-level `module_registry.cpp` in `generated/` root
- [ ] Entrypoint emits static library CMake target
- [ ] DLL modules emit shared library CMake targets with `REX_MODULE_EXPORT` on `ReXModule_Register`
- [ ] `--target` flag for selective module codegen (entrypoint always included)
- [ ] Legacy single-binary config backwards compatible (detected by absence of `[entrypoint]`)

### Runtime Module Loading
- [ ] `KernelState::RegisterRecompiledModule` stores `RecompiledModuleInfo` for runtime matching
- [ ] `LoadUserModule` matches normalized guest path, loads shared library, resolves `ReXModule_Register`
- [ ] Failed load (no matching recompiled module) warns and returns `X_STATUS_OBJECT_NAME_NOT_FOUND`
- [ ] DllMain called with `DLL_PROCESS_ATTACH` on load (`r3`=hmodule, `r4`=1, `r5`=0)
- [ ] DllMain called with `DLL_PROCESS_DETACH` on unload (`r3`=hmodule, `r4`=0, `r5`=0)
- [ ] Entry point address extracted from XEX header
- [ ] Shared library closed via `FreeLibrary`/`dlclose` after unregistration
- [ ] Module unload: decrement load count, remove from `user_modules_` when count reaches 0

### CLI
- [ ] `rexglue init` remains as alias for `rexglue init project`
- [ ] `rexglue init project` scaffolds manifest-based project (manifest, entrypoint config, CMake, presets, main.cpp, app header, rexglue.cmake)
- [ ] `rexglue init module <binary>` adds DLL module config, registers in manifest, updates rexglue.cmake
- [ ] `rexglue codegen` accepts manifest TOML or legacy single-binary config

### Out of Scope

- **Multi-XEX title switching** - games that launch secondary XEXs require full guest restart with overlapping address spaces. Needs kernel-level `TerminateTitle` + relaunch.
- **Parallel codegen** - per-binary pipelines are independent but deferred. Orchestrator loop has no shared mutable state to allow this later.
- **DLL dynamic relocation** - Xbox 360 does not relocate DLLs; base addresses are linker-assigned and must not collide.

### Modified Dependencies

No new external dependencies. Changes are internal to:
- `src/codegen/`, `include/rex/codegen/` - pipeline orchestration, hash construction, writer changes
- `src/system/`, `include/rex/system/` - FunctionDispatcher, Memory, KernelState, XModule
- `src/kernel/xboxkrnl/` - XexLoadImage/XexUnloadImage flow changes
- `src/rexglue/commands/` - CLI init and codegen command updates
- `include/rex/ppc/context.h` - new `last_indirect_target` field, updated `PPC_LOOKUP_FUNC` and `PPC_CALL_INDIRECT_FUNC` macros

### Acceptance Criteria

- [ ] Multi-binary project declared via manifest TOML builds successfully
- [ ] Codegen produces shared libraries for DLL modules and static library for entrypoint
- [ ] No conflicts with kernel export variables or existing allocations in `0x80000000+`
- [ ] `XexLoadImage` dynamically loads matching recompiled shared library and registers functions
- [ ] `XexUnloadImage` invalidates hash table entries + thunks, calls DllMain `DLL_PROCESS_DETACH`, closes shared library
- [ ] DLL reload after unload works correctly (function registrations swapped, new thunks allocated)
- [ ] `XexGetProcedureAddress` allocates thunks from pre-assigned range, resolves through hash table
- [ ] Cross-module indirect calls resolve correctly through the dispatch
- [ ] Calling an unloaded module function hits trap with diagnostic message showing guest address
- [ ] Single-binary projects continue to work unchanged (backwards compatible)
- [ ] `rexglue init project` / `rexglue init module` scaffold correctly
- [ ] `rexglue codegen --target` builds only specified modules plus entrypoint

### Related

_No response_

tomcl7:
# v0.8.0 - Multi-Binary DLL Support

## Summary

Large redesign of the runtime around multi-binary projects. Manifests now declare a single `[entrypoint]` (recompiled and linked into the host EXE, as before) and any number of `[[modules]]` (additional guest binaries built as shared libraries .dll/.so that `KernelState` registers, loads, and unloads at runtime). Each module owns its own `FunctionDispatcher`. The CLI is reworked around a `project_manifest.toml` manifest, with `codegen` replacing `migrate` and an upgrade workflow taking the place of the prior one shot path. `rex::system` and `rex::kernel` collapse into a single `rexruntime` SHARED target so cvars and dispatch can live behind a stable DLL boundary.
> [!CAUTION]
> This is a breaking release. Migration steps are at the bottom.

### Runtime

- Add `Runtime::Setup(PPCImageInfo, RuntimeConfig)` overload for module aware init
    - `PPCImageInfo` carries the per module function table, the `register_modules` callback, and the rexcrt heap opt in
    - `Runtime::Setup(RuntimeConfig)` still works for callers that don't need module wiring
- Make `FunctionDispatcher` per module with explicit register/unregister
    - Each loaded module owns its dispatch table; `KernelState` mediates registration
    - Replaces the previous single global table
- Change `FunctionDispatcher::AllocateThunk` signature to take `(PPCFunc*, uint32_t caller_address)`
    - Thunks are allocated from the calling module's reserve
    - Pass `ctx.lr` from inside guest code, or `0` for host initiated calls (routed to the entrypoint pool)
- Load and unload recompiled modules through `KernelState`
    - `KernelState::LoadUserModule` resolves guest paths against a registry populated by each module's `register_modules` callback
    - Manual `LoadLibrary` style loading of recompiled binaries is gone
- Switch `cvar` storage to accessor functions
    - `REXCVAR_GET(name)` for reads inside the same module
    - `REXCVAR_QUERY(type, name)` for cross module reads
    - Direct access to the storage variable that previously backed a cvar is no longer supported
    - Required for cvars to work across DLL boundaries
- Add platform agnostic shared library loader in `rex::system`
- Add `rex::system::NormalizeGuestPath` helper for canonical guest path form
    - Single canonical form for `game:\`, `update:\`, etc. across module registration and VFS lookups
    - Stricter than the old per call site handling: leading slashes, drive aliases, and case folding are now consistent
- Consolidate `Runtime::Setup` error paths and add a re-init guard

### Codegen

- Add manifest based project recompiler
    - Codegen consumes a `project_manifest.toml` listing every guest binary, target name, and per module config
    - Replaces the per-XEX invocation flow
- Add TOML parser for the manifest format
- Inline binary configs in the manifest format
- Stop emitting per call site unresolved-call warnings during codegen
    - The inline warnings now go to debug

### ReXGlue CLI

- Replace `rexglue migrate` with `rexglue codegen` plus an upgrade workflow
    - Run `rexglue codegen <legacy_config.toml>` to upgrade to a manifest based project
    - Manifest changes are detected automatically and surfaced through `migration_scan` and `upgrade_prompt`
    - No separate migration step to invoke
- Rework `rexglue init` for manifest first project layout
    - Top level `init` scaffolds a project anchored on a `game_root` directory
    - Discovers guest DLLs underneath `game_root` instead of taking a single XEX
    - Older single XEX manifests need to be regenerated
- Add `rexglue init module` subcommand for per module scaffolding inside an existing project
- Consolidate migration scanners
    - New identifier and call site scans feed the upgrade flow
- Rework codegen upgrade plan UX
- Tighten `init` overwrite gating
    - All writes are atomic
- General CLI overhaul (subcommand layout, help text, error paths, and fancy unicode glyphs)

### Build / CMake

- Narrow `rexglue_configure_target` to host applications and add `rexglue_configure_module_target` for guest DLL modules
    - `rexglue_configure_target(<target>)` keeps the same name but no longer applies cross-cutting flags or covers guest DLL output
    - `rexglue_configure_module_target(<target> [HOST <host_target>])` configures recompiled guest DLLs and colocates their output next to the host
    - `rexglue_apply_target_settings(<target>)` is the new home for the platform and compile flags both target shapes share
    - Whole-archive linking of `rex::kernel` is gone, since `rexruntime` is SHARED and exports kernel hooks directly
- Merge `rex::system` and `rex::kernel` into a single `rexruntime` SHARED library
    - Individual subsystems also consolidated into `rexruntime`
    - `rex_system`, `rex_kernel`, and the per subsystem targets are no longer linked separately
    - Link against `rexruntime` instead
- Refresh consumer DLL discovery and link guest modules to the host
- Exclude `REXGLUE_PROFILE_GUEST_FUNCTIONS` from test compile flags

### UI / Graphics

- Gate runtime launch on `game_data_root` with a fatal system dialog
    - Replaces the silent fallback to `<exe_dir>/assets` when the path is unset or invalid
- Move `video_mode` cvars from `xboxkrnl_video` to `ui/window`
    - Cvar names unchanged, `REXCVAR_GET` consumers unaffected
    - Code that declared or read them from inside the `xboxkrnl_video` translation unit needs to include `rex/ui/window.h`
- Mark D3D12 heap properties `inline constexpr`

### SDK Cleanup

- Remove deprecated `game_directory` positional argument from `ReXApp`
    - Set `game_data_root` from the appropriate `ReXApp` path resolution hook (`OnFinalizePaths`) before runtime launch
    - Or pass `--game_data_root=...` on the command line
    - Cvar fallback to `<exe_dir>/assets` still applies
- Add `CODEOWNERS`
- Remove deprecated macros alias
    - Remove legacy `PPC_*` aliases from `include/rex/hook.h`. Use `REX_HOOK` and friends
        - `PPC_HOOK`
        - `PPC_STUB`
        - `PPC_STUB_LOG`
        - `PPC_STUB_RETURN`
    - Remove legacy `PPC_*` aliases from `include/rex/ppc/context.h`. Use `REX_*` and `rex::ppc::kRound*`
        - `PPC_FUNC`
        - `PPC_FUNC_IMPL`
        - `PPC_EXTERN_FUNC`
        - `PPC_EXTERN_IMPORT`
        - `PPC_WEAK_FUNC`
        - `PPC_JOIN`
        - `PPC_XSTRINGIFY`
        - `PPC_STRINGIFY`
        - `PPC_ROUND_NEAREST`
        - `PPC_ROUND_TOWARD_ZERO`
        - `PPC_ROUND_UP`
        - `PPC_ROUND_DOWN`
        - `PPC_ROUND_MASK`
    - Remove legacy `PPC_CONFIG_*` configuration defines. Use the matching `REX_CONFIG_*`
        - `PPC_CONFIG_NON_ARGUMENT_AS_LOCAL`
        - `PPC_CONFIG_NON_VOLATILE_AS_LOCAL`
        - `PPC_CONFIG_SKIP_LR`
        - `PPC_CONFIG_CTR_AS_LOCAL`
        - `PPC_CONFIG_XER_AS_LOCAL`
        - `PPC_CONFIG_RESERVED_AS_LOCAL`
        - `PPC_CONFIG_SKIP_MSR`
        - `PPC_CONFIG_CR_AS_LOCAL`
    - Remove `PPC_*` aliases from codegen output (`init_h.inja`, `ppc_config_h.inja`). Recompile to regenerate against the `REX_*` names
        - `PPC_FUNC_PROLOGUE`
        - `PPC_CALL_FUNC`
        - `PPC_LOOKUP_FUNC`
        - `PPC_CALL_INDIRECT_FUNC`
        - `PPC_SET_FLUSH_MODE`
        - `PPC_UNIMPLEMENTED`
        - `PPC_QUERY_TIMEBASE`
        - `PPC_CHECK_GLOBAL_LOCK`
        - `PPC_ENTER_GLOBAL_LOCK`
        - `PPC_LEAVE_GLOBAL_LOCK`
        - `PPC_PHYS_HOST_OFFSET`
        - `PPC_RAW_ADDR`
        - `PPC_LOAD_U{8,16,32,64}`
        - `PPC_LOAD_STRING`
        - `PPC_STORE_U{8,16,32,64}`
        - `PPC_MEMORY_SIZE`
        - `PPC_IS_MMIO_ADDR`
        - `PPC_MM_LOAD_U{8,16,32,64}`
        - `PPC_MM_STORE_U{8,16,32,64}`

## Misc Fixes

- Scope `AppManager` dispatch iterators so debug iterator validation doesn't trip on shutdown

## Migration Guide

- Run `rexglue codegen <project.toml>` directly to regenerate guest project files
    - Run the CLI itself, not the cmake util target. The upgrade detector is interactive and prompts for consent before applying changes, which a build time target can't surface
    - Manifests with the old single XEX layout need to be regenerated against the new `game_root` anchoring
- Update CMake projects
    - Keep `rexglue_configure_target` on the host EXE; it no longer applies cross-cutting flags on its own
    - Call `rexglue_configure_module_target(<target> HOST <host_target>)` on each recompiled guest DLL
    - Add `rexglue_apply_target_settings(<target>)` anywhere you previously relied on `rexglue_configure_target` to set platform and compile flags
    - Link against `rexruntime` instead of `rex_system`, `rex_kernel`, or the per subsystem targets
- Replace `PPC_*` macros with `REX_*` in any manual hook code
    - Codegen does this for generated files
    - Same sweep for `PPC_CONFIG_*` defines if you set any in build flags
- Replace `--game_directory` invocations
    - Set `game_data_root` from the appropriate `ReXApp` path resolution hook (`OnFinalizePaths`)
    - Or pass `--game_data_root=...` on the command line
- Route through `KernelState` anywhere code touched the global `FunctionDispatcher` directly
    - `Runtime::Setup` callers are unaffected
- Add a `caller_address` argument to any custom `FunctionDispatcher::AllocateThunk` call
    - `ctx.lr` from inside guest code, or `0` for host initiated calls
- Switch cvar reads off the backing storage variable
    - `REXCVAR_GET(name)` for reads inside the same module
    - `REXCVAR_QUERY(type, name)` for cross module reads
- Switch any manual guest path canonicalization to `rex::system::NormalizeGuestPath`
    - Manually constructed paths that relied on the prior leniency (extra leading slashes, drive alias casing) may need to be adjusted


Loreaxe (commented):
Works great on Linux once you get your head around the new codegen and migration.

Loreaxe (dismissed):

SolarCookies:
Worked great for retip, previous versions of rexglue since 0.7.4 where having issues with retip. Great work! It took a minute to  get the new code gen up and running but that's mostly user error and having to stub stuff from my personal fork. Unfortunately Viva Pinata does not have multi binaries so I'm not able to test the functionally of that. Ill be sure to start making PRs in the near future to move over some of the stuff I have on my fork that would help other projects like the shader debugger and well.. Grain beat me too it with the texture injection.

SolarCookies (approved):

Mystixor (dismissed):

Graine25 (dismissed):

src/rexglue/commands/codegen_command.cpp
```diff
@@ -2,39 +2,421 @@
...
+      return {"Deleted", "delete"};
+  }
+  return {"Touched", "?     "};
+}
+
+bool ApplyEntry(const OverwriteEntry& entry) {
+  std::error_code ec;
+  switch (entry.action) {
+    case OverwriteAction::Write: {
+      fs::create_directories(entry.path.parent_path(), ec);
```
davesilva: We should skip `create_directories` if `entry.path.parent_path()` is an empty string.  Currently the migration fails here if you're already in the project directory and run `rexglue codegen game_config.toml`.

tomcl7: Good catch. Fixed with [2bdb97f](https://github.com/rexglue/rexglue-sdk/pull/335/commits/2bdb97f95f154f32d281aaa08446ae007b8ca117)

davesilva (approved):

PR #344, opened 2026-05-23
Fix Daytona USA XBLA rendering issues and other improvements

Subarasheese:
These are changes I had to make in order to make Daytona USA XBLA playable with graphics being rendered correctly.

### 1. Added Vulkan DXT5 RGBA decode flag and safer SPIR-V interpolator handling

* Added a new graphics flag: `vulkan_force_dxt5_rgba_decode`.
* Updated the SPIR-V translator to check whether each interpolator is available before loading it.
* If an interpolator is unavailable, the shader now falls back to a default `float4(0)` value.
* This improves robustness around fragment shader input interpolation.

### 2. Added Xbox 360 `c255` helper constant handling and corrected texture fetch exponent bias source

* Added special handling for Xbox 360 shader constant `c255`.
* `c255` is treated as a fixed helper constant with values equivalent to `float4(0, 1, 0, 1)`.
* Updated texture fetch processing to load fetch constant **word 3** for result exponent biasing.
* Corrected exponent bias extraction to use bits `3:18` from fetch constant word 3 instead of word 4.

### 3. Improved Vulkan texture decode behavior for DXT5 and `k_8` gamma-swizzled textures

* Added handling for `k_8` textures using the `XYZ1` swizzle pattern with unsigned gamma signs.
* RGB channels are decoded as gamma-encoded while alpha remains linear for this case.
* Added the `vulkan_force_dxt5_rgba_decode` GPU flag:

  * Forces DXT4/5 textures through the RGBA8 compute decode path.
* Updated Vulkan texture format selection so DXT5 can fall back to shader-decoded RGBA8 when:

  * the new force-decode flag is enabled, or
  * BC3/DXT5 linear filtering support is unavailable.

### 4. Hardened `XamShowMessageBoxUI` and implemented shared message box handling

* Refactored guest UTF-16 string loading into a helper function.
* Added safer handling for null title, text, and button pointers.
* Split message box behavior into a shared `XamShowMessageBoxUICommon` path.
* Skips null button pointers and ignores empty button labels.
* Adds a default `"OK"` button when no valid buttons are provided.
* Clamps the active button index to the available button list.
* Added null checks before writing the selected button result back to guest memory.
* Updated both UI and headless paths to avoid writing through a null `result_ptr`.
* Implemented `XamShowMessageBoxUIEx_entry` by forwarding it through the shared common message box handler instead of leaving it as a stub.

### 5. Improved title termination thread lifetime safety

* Updated suspended thread tracking during title termination to store retained `XThread` object references instead of raw pointers.
* Suspended threads are now retained before the global lock is released.
* This should reduce the risk of a suspended thread object becoming invalid while termination is being performed outside the lock.


tomcl7 (changes requested):
Can you redo the commits to match the work? Also please run clang-format on the project to format your changes.

Graine25:
I think these changes are too significant to be bundled in one pull request and no issues explaining the causality. Let's circle back on this once we understand the problem + solutions

PR #346, opened 2026-05-23
Kameo aarch64 + Pocket DS port enablement

Azkali:
## Summary

Branch carries the SDK changes that took Kameo from "amd64-only" to running on Tegra X1 (Switch) and Snapdragon SM8550 / Adreno 740 / Mesa Turnip (Ayaneo Pocket DS), plus a handful of amd64-applicable fixes that fell out along the way.

Note: branch is 69 commits behind upstream `main` (forked off v0.7.6, upstream is now at v0.8.0). Marked draft pending rebase — happy to do that on request, or to split into smaller topical PRs if you'd prefer cherry-picking.

### What's in here

**Codegen / PPC correctness**
- `fix(codegen): emit memory fences for sync/lwsync/eieio` — required for the ARM memory model.
- `fix(codegen): NULL-guard REX_CALL_INDIRECT_FUNC against IP=0 cascade` — without this, an indirect-NULL on aarch64 cascades into a 1.3 M/s SEGV loop.
- `fix(ppc): re-mask FP exceptions in FPSCRRegister::setcsr` + `fix(ppc): use Platform::InitHostExceptions in setcsr for arch-correct polarity` — Tegra X1 / Adreno hosts trip FE_INVALID on guest float ops without these.
- `feat(codegen): multi-pass gap-fill for post-tail-call thunks`.

**aarch64 / SEH / memory**
- `fix(seh): catch SIGBUS on aarch64; chain to original disposition when unhandled`.
- `fix(audio): make SDL audio init failure non-fatal for headless boot` — lets the runtime come up on hosts where SDL audio init returns a DUMMY backend.

**Bink cinematic overlay (FFmpeg + SDL3)**
- `feat(system): Bink cinematic overlay backend (FFmpeg + SDL3)` — libavformat/libavcodec/libswscale via dlopen-RTLD_LOCAL so the dynamic libavcodec doesn't shadow the statically-linked XMA codec.
- `feat(ui): BinkUIDrawer — composite cinematic overlay over swap`.
- `feat(kernel): notify Bink overlay on guest .bik file opens` — `NtCreateFile_entry` → `BinkOverlay::OnFileOpened` wiring.
- `fix(system/bink): read width/height from codecpar, not codec_ctx` — libavcodec 62.x on Fedora 44 aarch64 returns swapped dims via `codec_ctx` for binkvideo after `avcodec_open2`, producing half-black / half-striped frames.

**Pocket DS replay fixes (2026-05-13/14)**
- `feat(ui/vulkan): UpdateTextureData for streaming re-uploads` — Bink overlay was allocating a fresh VkImage per frame and fragmenting Mesa Turnip's GPU pool; this lets BinkUIDrawer re-upload pixels in place. New entries in the pending-upload queue carry a `reupload` flag so the pre-transfer barrier transitions from `SHADER_READ_ONLY` rather than `UNDEFINED`.
- `fix(ui/gtk): marshal off-thread RequestPaint via g_idle_add + debounce` — `gtk_widget_queue_draw` must run on the GTK main thread but engine threads call `RequestPaintImpl` directly; on Plasma 6 KWin Wayland / Xwayland the off-thread call didn't reliably wake the loop, leaving the screen frozen while `PRESENT` kept logging. Atomic-flag debounce coalesces concurrent requests so present-rate calls don't take a GLib mutex + idle-source allocation each time.

**Input**
- `fix(input): drop has_focus_ gate in MnK OnKeyDown` and `fix(input/mnk): drop has_focus_ gate in GetState` — KDE Wayland never raises X focus on the embedded window, so the gate dropped every key. (Also fine on X11.)

**Save state / data root**
- `feat(sys): mount user_data_root/saves as save:`.

**RenderDoc capture (opt-in)**
- `feat(ui): RenderDoc early init + capture-path/keybind config`.
- `feat(graphics/vulkan): auto-trigger RenderDoc capture via env var`.
- `feat(graphics/vulkan): bracketed RenderDoc capture (no-present games)` — needed for diagnosing 0-PRESENT Switch boots before the FMV cycle fixes landed.

**Misc**
- `feat(krnl): dump_swaps cvar to capture VdSwap source bytes`.
- `feat(graphics): invalidate_shared_memory_per_frame cvar`.

### Reverted experiments (still in history)

A handful of shared-memory-import experiments (`VK_EXT_external_memory_host` path, `EnableAccessCallbacks` stub, GPU `HostWriteBarrier`, etc.) live in the history as feat→revert pairs. Net diff is zero for those; if a rebase is wanted I can squash them out.

## Test plan

- [ ] amd64 / Fedora 44 / Mesa RADV: Kameo boots → disclaimer → splash → PRESS START → book menu → in-game (existing target, regression-tested).
- [ ] Switch (Tegra X1) / L4T: boots → up to 320 PRESENT in current head; FMV intro cycle is the bottleneck (separately tracked).
- [ ] Ayaneo Pocket DS / Adreno 740 / Mesa Turnip: PRESS START → book menu → Solo → Game List → "Start without saving" → Thorn's Castle Opening Scene end-to-end. Cinematic FMVs render via the Bink overlay.
- [ ] Rebase onto current upstream `main` (v0.8.0) before unmarking draft.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Graine25:
Closing this prematurely since this PR breaks the rules set in [CONTRIBUTING.md](https://github.com/rexglue/rexglue-sdk/wiki/Contributing), but I'll give my two cents on this.

We currently have someone we know working on Kameo: Elements of Power, and they have told us that they have not ran into similar issues on linux-amd64. We would prefer if you could write the problems out into a Github issue rather than submitting everything all in one Pull Request.

Additionally, it is very hard for us to read 5 different issues, so try to space things out and only solve 1 issue at a time. Everyone who contributes is doing this voluntarily so we cannot get to every single issue. Discussion in the Discord server (linked in the README) is highly recommended so we can figure out the issue rather than assuming it exists.

We can revisit all these issues overtime but try to space things out and personally (not just agentically 😄) make sure we've solved every issue

Azkali:
I understand thank you for your reply; I did have everything working and solved on Kameo; I can and will make more digest commits for you to review; Will also join your discord

Issue #348, opened 2026-05-27
[Bug]: Duplicate cvar registrations on `rexglue codegen` startup


birabittoh:
### Description

When running `rexglue codegen`, seven duplicate cvar registration errors are emitted to stderr before the tool header is printed. The errors appear because `rexglue` links `librexruntime.so` as a shared library, causing the runtime's static initializers (which register cvars) to execute twice: once when the dynamic linker loads the shared library, and once via the tool binary's own static-initialization path. The registrations are ignored (second instance is a no-op).

### Steps to Reproduce

Run the codegen tool: `sdk/bin/rexglue codegen kameorepowered_manifest.toml`

### Expected Behavior

Tool starts cleanly with no cvar errors:

```
ReXGlue v0.8.1.24 - Xbox 360 Recompilation Toolkit
  ...
```

### Actual Behavior

Seven duplicate-registration errors are printed before the header:

```
[error] [core] cvar: duplicate registration of 'log_level'; second registration ignored
[error] [core] cvar: duplicate registration of 'log_file'; second registration ignored
[error] [core] cvar: duplicate registration of 'log_verbose'; second registration ignored
[error] [core] cvar: duplicate registration of 'log_noisy'; second registration ignored
[error] [core] cvar: duplicate registration of 'log_flush_interval'; second registration ignored
[error] [core] cvar: duplicate registration of 'log_max_file_size_mb'; second registration ignored
[error] [core] cvar: duplicate registration of 'log_max_files'; second registration ignored
ReXGlue v0.8.1.24-dev.g0490a46 - Xbox 360 Recompilation Toolkit
  ...
```

### Platform

Linux

### Clang version

22.1.5

### Version / Commit

v0.8.1.24-dev.g0490a46

### Additional Environment Info

This bug doesn't affect Windows.

cybermobile:
Same root cause as #347 — this and the exit-time SIGSEGV there are two symptoms of `rexcore` (OBJECT lib) being embedded into both `librexruntime.so` and the `rexglue` exe. Posted a confirmed repro + a working `rexcore` OBJECT→SHARED fix over in #347 that clears these 7 duplicate registrations.

tomcl7:
Fixed https://github.com/rexglue/rexglue-sdk/commit/87afd28318cae94e8415049651277bd9033c28b3

PR #349, opened 2026-05-27
build: link rexruntime statically into rexglue

### LINKED ISSUE #347: [Bug]: codegen crashes with SIGSEGV on exit (exit code 139)
### Description

`rexglue codegen` completes its work successfully but terminates with a segmentation
fault during process shutdown. The codegen output is written correctly before the
crash, but the non-zero exit code (139) causes any script or CI step that checks
`$?` to treat the run as a failure.

The crash occurs in `rex::cvar::UnregisterFlag` while it is iterating an
`std::unordered_map` that has already been destroyed. `FlagRegistrar` static objects
inside `librexruntime.so` have their destructors invoked by `__cxa_finalize` during
`librexruntime.so` teardown. By that point, the function-local statics backing the
cvar registry (in the main `rexglue` executable) have already been destroyed, leaving
dangling references. The backtrace confirms this:

```
#4  rex::cvar::UnregisterFlag          (rexglue)
#5  rex::cvar::FlagRegistrar::~FlagRegistrar  (rexglue)
#6  __cxa_finalize                     (libc.so.6)
#7  __do_global_dtors_aux              (librexruntime.so)
```

### Steps to Reproduce

```bash
rexglue codegen <any_manifest.toml>
echo $?
```

### Expected Behavior

The process exits with code `0` after writing the generated sources.

### Actual Behavior

Codegen completes and writes all files, then crashes:

```
  ✓ default.xex (4.8s)
Done in 4.9s.
Errore di segmentazione    (core dump creato)
$ echo $?
139
```

### Platform

Linux

### Clang version

22.1.5

### Version / Commit

0.8.1.24-dev.g0490a46

This bug doesn't affect Windows.

### Related

_No response_

birabittoh:
Introduce rexruntime_objects (OBJECT library) compiled from the system + kernel sources, and rexruntime_static (STATIC) that absorbs it alongside all runtime deps. rexcodegen now links rexruntime_static instead of the shared library, so rexglue becomes a fully self-contained executable with no librexruntime.so dependency at runtime.

The shared rexruntime target and its public install/export interface are left unchanged, so host applications continue shipping the shared library exactly as before.

Removes the explicit rexruntime link from unit_tests to avoid mixing the shared and static copies of the runtime in the same executable (the runtime now arrives through rexcodegen -> rexruntime_static). ppc_tests is unchanged as it does not link rexcodegen.

Tested working on both Windows 11 and Arch Linux (CachyOS).

Fixes #347 and #348.

### DIFF
--- a/src/codegen/CMakeLists.txt
+++ b/src/codegen/CMakeLists.txt
@@ -118,7 +118,7 @@ target_include_directories(rexcodegen
 target_link_libraries(rexcodegen
     PUBLIC
         rexcore
-        rexruntime
+        rexruntime_static
         fmt::fmt
         spdlog::spdlog
     PRIVATE
--- a/src/kernel/CMakeLists.txt
+++ b/src/kernel/CMakeLists.txt
@@ -64,6 +64,12 @@ set(REXKERNEL_SOURCES
 )
 
 list(TRANSFORM REXKERNEL_SOURCES PREPEND "${CMAKE_CURRENT_SOURCE_DIR}/")
-target_sources(rexruntime PRIVATE ${REXKERNEL_SOURCES})
+target_sources(rexruntime_objects PRIVATE ${REXKERNEL_SOURCES})
 
-target_link_libraries(rexruntime PRIVATE o1heap rexcore rexfilesystem rexgraphics rexui rexaudio rexinput imgui)
+set(_rex_kernel_deps o1heap rexcore rexfilesystem rexgraphics rexui rexaudio rexinput imgui)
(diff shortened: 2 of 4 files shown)


tomcl7:
I do not understand this PR. You are wanting to change the shared lib back into a static lib to work around other issues you are experiencing? The solution should be tracking and fixing the issues, not reverting intentional architectural changes.

Please explain this a bit more and disclose if you created this with the assistance of an LLM.

birabittoh:
Closing because there are clearly better approaches to those issues.

Issue #338, opened 2026-05-11
[Bug]: vaddsws emits per-byte blend, corrupts output without real overflow


ItsNotPaths:
### Description

`build_vaddsws` in `src/codegen/builders/vector.cpp` (`Release v0.7.7`, `a9bf7a2`) selects between `sum` and the saturation value using `simde_mm_blendv_epi8` (per-byte selection). PPC `vaddsws` is a per-32-bit-word saturating add; the overflow-detection mask `(a^sum) & (b^sum)` is meaningful only at bit 31 of each dword, and bits 0–30 are arbitrary. Per-byte blending also samples bits 7, 15, and 23 of the mask, so bytes of `sum` get replaced by bytes of `sat_val` (`0x7FFFFFFF` / `0x80000000`) even when no real overflow has occurred.

Reference behavior — see xenia's `x64_seq_vector.cc:707-741` `VECTOR_ADD INT32_TYPE` saturate path — is per-dword blending via `vblendvps`.

### Steps to Reproduce

Drop into a standalone `.cpp` and compile against `simde`:

```cpp
#include <simde/x86/sse4.1.h>
#include <cstdio>
#include <cstdint>

int main() {
    // Two positive int32s well within range — sum cannot overflow.
    alignas(16) int32_t in[4] = { 0x00800000, 0x00800000, 0x00800000, 0x00800000 };

    simde__m128i a = simde_mm_load_si128((simde__m128i*)in);
    simde__m128i b = simde_mm_load_si128((simde__m128i*)in);
    simde__m128i sum = simde_mm_add_epi32(a, b);
    simde__m128i overflow = simde_mm_and_si128(
        simde_mm_xor_si128(a, sum),
        simde_mm_xor_si128(b, sum));
    simde__m128i sat_val = simde_mm_xor_si128(
        simde_mm_srai_epi32(a, 31),
        simde_mm_set1_epi32(0x7FFFFFFF));

    alignas(16) int32_t out[4];
    simde_mm_store_si128((simde__m128i*)out,
        simde_mm_blendv_epi8(sum, sat_val, overflow));

    std::printf("got      = 0x%08X\n", out[0]);
    std::printf("expected = 0x%08X\n", 0x01000000);
}
```

[...]

PR #351, opened 2026-05-30
fix(codegen): make vaddsws use blendv_ps via casts in vaddsws

Graine25:
Before, `vaddsws` used `blendv_epi8`, so the overflow mask could accidentally select saturation bytes based on bits inside each 32-bit lane, corrupting results even when no signed word overflow occurred. After the change, the selection is done per 32-bit lane with `blendv_ps`, so each word either keeps the exact add result or is fully replaced with the correct signed saturation value.

Titles like Banjo Kazooei: Nuts & Bolts, KAMEO: Elements of Power, and Forza Horizon 1 all contained this instruction and resulted in models jittering/resizing themselves when they were not supposed to. The fixes suggested in commit `fix(codegen): make vaddsws use blendv_ps via casts in vaddsws` addresses these issues.

Solves #338


tomcl7 (approved):
Looks good.

Issue #235, opened 2026-03-16
[Feature]: Mac Support


Graine25:
### Summary

Enable building and running ReXGlue projects on macOS. The repo already has Unix scaffolding, but macOS is not yet a supported end-to-end target, which blocks native development and testing on Apple hardware.

### Proposed Solution

Build on the Linux/Arm64 work and make a Mac Port

### Key Features:
- MoltenVK rendering
- Proper mac-arm64 cmake support
- Cocoa windowing/runtime support
- CI/Validation scripts to make sure C++ codebase is darwin compliant

### Build System
- [ ] macOS configure/build presets
- [ ] Apple-specific install/runtime staging
- [ ] Apple Silicon support

### Runtime
- [ ] Cocoa window/context implementation
- [ ] MoltenVK-backed Vulkan surface creation
- [ ] End-to-end sample app launch on macOS

### Quality
- [ ] Basic macOS test coverage
- [ ] Documentation for setup and known limitations

### Out of Scope

- Native Metal backend: Metal 4 lacks support for a lot of features that D3D12 supplies for the gpu backend

### Modified Dependencies

- MoltenVK - Vulkan implementation on macOS
- AppKit / Foundation / QuartzCore - macOS windowing and surface integration in CMake

### Acceptance Criteria

- [ ] A macOS user can configure and build the SDK from a clean checkout
- [ ] A generated or sample app opens a native macOS window and renders successfully
- [ ] Vulkan works through MoltenVK without requiring ad hoc local fixes
- [ ] The project doesn't hinder performance or affect windows and linux

PR #356, opened 2026-06-11
[Feature] macOS Support

Graine25:
This PR adds end-to-end macOS support to rexglue-sdk, covering everything from CMake build presets and CI pipelines to the windowing layer and MoltenVK-backed Vulkan rendering. It introduces Darwin-specific backends for core systems (memory, threading, filesystem, exception handling, clock) alongside a brand new SDL/Cocoa windowing layer and MoltenVK loader that all together replace the Linux/Windows-only assumption in the runtime. The goal is a clean first-class macOS target — both arm64 and amd64 — that builds from a fresh checkout, opens a native window, and renders through MoltenVK without any local workarounds.

#### Core (`src/core/`)

This is the foundation of the port. Five Darwin-specific translation units were introduced in `d3e7bf6` and then individually hardened in follow-up commits:

- **`exception_handler_mac.cpp`** (`9a8e2aa`) — Full Darwin exception handler. Maps `mcontext_t` register fields to the emulator's abstract register file and adds `SIGBUS` handling — macOS raises `SIGBUS` for invalid memory access where Linux raises `SIGSEGV`.
- **`filesystem_mac.cpp`** (`3897b24`) — Swaps out Linux-only `readlink("/proc/self/exe")` for `_NSGetExecutablePath` and replaces `fseeko64`/`ftello64` (glibc extensions) with standard POSIX `fseeko`/`ftello`, which is what Darwin actually exposes.
- **`mapped_memory_mac.cpp`** (`9cf42b5`) — Same glibc-extension problem as the filesystem layer: `fstat64`/`ftruncate64` don't exist on Darwin. Swapped for POSIX `fstat`/`ftruncate`.
- **`memory_mac.cpp`** (`c96f117`) — Mach VM backend. Uses `mach_vm_query_region` for memory range queries, `MAP_JIT` for JIT-writable anonymous mappings (required by Hardened Runtime on Apple Silicon), and clamps `shm_open` names to Darwin's 32-character limit.
- **`threading_mac.cpp`** (`8fa4f83`) — Replaces Linux `pthread`/`SIGUSR1` suspend-resume with `mach_thread_suspend`/`mach_thread_resume`, derives thread IDs from `mach_thread_self()`, and stubs CPU affinity masks since macOS has no first-class equivalent.

SDL window drawing has also been enabled:
- **`window_mac.cpp`** + **`windowed_app_context_mac.cpp`** + **`windowed_app_main_mac.cpp`** (`60bd157`) — The entire Cocoa windowing layer, written from scratch (~740 lines). `window_mac.cpp` owns the SDL Metal window lifecycle, the Cocoa run loop, input dispatch, and the `SDL_EVENT_QUIT` path. `windowed_app_context_mac.cpp` is the platform-specific app context that wires the run loop into the ReX app lifecycle. `windowed_app_main_mac.cpp` is the entry-point shim — it initializes the context, calls `RunMainLoop()`, and hard-exits via `std::_Exit` on return to prevent a deadlock when Cmd+Q fires `SDL_EVENT_QUIT`, which skips `OnClosing` and would otherwise block `InvokeOnDestroy` on kernel threads that are no longer draining. 

The debugger was also split into `dbg_mac.cpp` (`056905d`), which wires into Darwin's `PT_DENY_ATTACH` anti-debug facility and reads the debug flag through `sysctl` the way macOS expects.

---

#### GPU / Graphics (`src/ui/vulkan/`, `src/graphics/`)

Vulkan on macOS runs over MoltenVK, which has constraints the Linux/Windows path never sees:

- **`vulkan_moltenvk.cpp/.h`** (`2b54711`) — New translation unit that dlopen-loads `libMoltenVK.dylib` at runtime (it doesn't ship as a system library), grabs `vkGetInstanceProcAddr` through the loaded handle, and forwards the Metal layer pointer from SDL.
- **`vulkan_instance.cpp`** (`a1911ac`, `2b54711`) — `VK_EXT_metal_surface` is conditionally requested under `__APPLE__`. The extension function is resolved through the MoltenVK loader rather than the ICD.
- **`vulkan_presenter.cpp`** (`a1911ac`, `60bd157`) — Added the `VkMetalSurfaceCreateInfoEXT` path for surface creation, pulling the `CAMetalLayer` from SDL.
- **`vulkan_device.cpp`** (`12bd421`) — MoltenVK doesn't support `geometryShader` or `fillModeNonSolid`. Both cvars now default to `false`. Without this the device selection loop rejected every GPU at startup and the process exited silently.
- **`spirv_translator.cpp`** (`047b453`) — The translator was dropping `OpDecorateString` for rectangle-style outputs, which MoltenVK's SPIRV-Cross pipeline rejected. Fixed to preserve those outputs through the translation pass.

---

#### Kernel / System (`src/system/`, `src/kernel/`)

- **`xmemory.cpp`** (`66dff88`) — Adds macOS-specific `vm_protect` calls for protection changes and `mach_vm_region_recurse` for querying a given address's protection flags. Referenced xenia-edge's implementation at `993a676`.
- **`xam_net.cpp`** / **`xboxkrnl_threading.cpp`** (`cda459e`) — Several `#ifdef` guards that were Linux/Windows-only now include `__APPLE__` so the stubs compile without falling into Linux-assumed `#else` branches.
- **`renderdoc_api.cpp`** (`44457a7`) — RenderDoc's `dlopen` path differs between Linux and macOS. Added a Darwin branch so the `RENDERDOC_GetAPI` lookup doesn't hard-fail on startup when no RenderDoc is present.

---

### Changes Made to Existing Files

- **`CMakePresets.json` + `CMakeLists.txt` + `thirdparty/CMakeLists.txt`** (`bef2683`, `99aa036`) — Added `mac-arm64-relwithdebinfo`, `mac-arm64-release`, and `mac-amd64-release` presets. The root `CMakeLists.txt` got an `if(APPLE)` branch that brings in `OBJC`/`OBJCXX` languages and sets `MACOSX_DEPLOYMENT_TARGET`. The `thirdparty/` `IS_ARM64`/`IS_X86_64` detection was also fixed here — it was reading `CMAKE_SYSTEM_PROCESSOR`, which is always `arm64` on Apple Silicon even when cross-compiling for x86_64. The fix checks `CMAKE_OSX_ARCHITECTURES` first (the authoritative field on Apple) and only falls back to `CMAKE_SYSTEM_PROCESSOR` on non-Apple hosts. 

- **`cmake/rexglue_helpers.cmake`** + **`cmake/rexglue_install.cmake`** (`4366ab8`, `2b54711`) — Added an `elseif(APPLE)` branch in `rexglue_configure_target` so consuming projects pick up `windowed_app_main_mac.cpp` instead of the POSIX entrypoint (which includes GTK3 headers — without this any macOS consumer failed with `'glib.h' file not found`). The install rule was extended to package `windowed_app_main_mac.cpp` alongside the Windows and POSIX entrypoints. The same helpers file also gained `rexglue_find_macos_vulkan_runtime` and `_rexglue_copy_macos_vulkan_runtime`, which locate and stage `libMoltenVK.dylib`, `libvulkan.1.dylib`, and `MoltenVK_icd.json` next to the host binary at build time. The probe order is: `REXGLUE_VULKAN_RUNTIME_DIR`, `REX_VULKAN_SDK`, `VULKAN_SDK`, `~/VulkanSDK/*`, then Homebrew prefixes.

- **`.github/workflows/_build-platform.yaml`** (`59b5fb4`, `98e57fc`) — Added `mac-amd64` and `mac-arm64` runner entries (both on `macos-26`; earlier runners don't have `std::from_chars` float support in Apple libc++). A "Set compilers (macOS)" step now prepends Homebrew LLVM's `bin` to `$GITHUB_PATH` and sets `CC`/`CXX` before CMake runs. This was needed because `cmake --preset` resolves `clang` via PATH, which on a stock macOS runner hits `/usr/bin/clang` (AppleClang) first.

- **`src/core/clock_posix.cpp`** + **`src/core/fiber_posix.cpp`** (`e72859e`) — Two existing POSIX files that needed Darwin fixups. `clock_posix.cpp`: `clock_gettime(CLOCK_MONOTONIC_RAW)` returns nanoseconds while `host_tick_frequency_platform()` returned ~24 MHz (a mach tick rate), causing ~41× time acceleration at runtime. Fixed by switching `host_tick_count_platform()` to `mach_absolute_time()` on `__APPLE__` so count and frequency are in the same unit. `fiber_posix.cpp`: `ucontext_t` is deprecated since macOS 12 — added a `_XOPEN_SOURCE` suppression guard to silence warnings without pulling in the full POSIX compatibility shim.

- **`resources/templates/codegen/init_h.inja`** (`ee00f7d`) — `__attribute__((alias(...)))` is ELF-only. The codegen template was emitting it unconditionally, causing link errors on Darwin. An `#if defined(__ELF__)` guard was added around the alias declarations.

- **`thirdparty/crypto/sha256.cpp`** (`b132360`) — The SHA-256 round constants were initialized with function-call expressions in a static array, which triggers macOS's text-relocation restriction under Hardened Runtime. Rewritten to avoid relocatable references in read-only data.

- **`src/input/sdl/sdl_input_driver.cpp`** (`8baf6a9`) — SDL on macOS only processes events on the main thread. Calling `SDL_PollEvent` from a secondary thread crashes the Cocoa event pump. Added a main-thread dispatch guard to the input driver's poll loop.

---

### Concerns for Merging Upstream

**The `__attribute__((alias))` stub in `init_h.inja` (`ee00f7d`) is the main open question.** The commit message flags it directly: `alias` is ELF-only, and the current fix is a guard rather than a real solution. Any codegen path that relies on aliasing for zero-cost guest function dispatch just doesn't get it on Darwin. If that ever becomes load-bearing rather than a shortcut, the macOS build will silently fall back to a slower path with no warning. Needs a decision: either a Mach-O-compatible trampoline or a documented pragma that reviewers can grep for.

**`MAP_JIT` in `memory_mac.cpp` (`c96f117`)** requires the process to have the `com.apple.security.cs.allow-jit` entitlement under Hardened Runtime. The code allocates with `MAP_JIT` but there's no accompanying entitlement plist. This is fine for unsigned dev builds and CI runners, but Gatekeeper will reject any notarized distribution build without the entitlement. Distribution packaging will need to account for this before the port ships to end users.

https://github.com/user-attachments/assets/0ec14d48-500f-48b4-892c-6e1fa9b6ed31

Heavy testing is required for us to send something like this out, so I would like for more contributors to give insight on something like this, if you have access to a mac device. All feedback is welcome on a rather significant addition to the SDK. Solves #235


tomcl7:
Once the conflicts are resolved, have @mrcmunir and @Mystixor provide testing for Linux/Windows.

Thanks for putting this together.

Mystixor (approved):

mrcmunir (approved):

Graine25:
Rewriting

Issue #350, opened 2026-05-29
[Feature]: Achievement Support


Graine25:
### Summary

Add an in-runtime Achievements page/overlay for ReXGlue games. The current system has no achievement-related support and no user-facing page to view achievement names, descriptions, gamerscore, lock/unlock status, or function key suggested to make something like this viewable.

### Proposed Solution

Implement an ImGui-based Achievements overlay/page inside ReXGlue. The feature should reuse the SDK's existing XDBF achievement metadata support and XGI achievement-write path. Routing it directly from the binary similar to how Xenia-Edge does it is a strong path, but it locks developers out of being able to control the metadata themselves. We could implement a carg in ReXGlue that could dump information related to achievements out of an xex directly to an assigned folder, like so

`rexglue init --achievements /path/to/xex/ /path/to/generated/achievements/
`

After dumping metadata, we can enable a setting in ReXGlue which builds the metadata into a recomp and attaches it directly to an imgui window. Attach the F7 function key to our imgui window, akin to how the runtime settings currently works, that opens a simple window and allows users to scroll through the achievements of the recompilation.

### CLI / Metadata Generation
- [ ] Extend `rexglue init` with an achievements generation option.
- [ ] Accept an input XEX path and an output folder path for generated achievement metadata.
- [ ] Generate an editable achievements metadata file, such as `achievements.toml` or `achievements.json`.
- [ ] Optionally extract achievement icons/images if available.
- [ ] Handle games with no achievement metadata gracefully.
- [ ] Print clear CLI errors for invalid XEX paths, invalid output folders, or unsupported achievement data.

### UI / Overlay
- [ ] Add `AchievementsOverlayDialog` as an `ImGuiDialog` subclass.
- [ ] Add a keybind or menu toggle for opening the achievements page.
- [ ] Add the new files to the relevant CMake target/source list.
- [ ] Bind the overlay to `F7`.
- [ ] Match the style and behavior of the existing runtime settings/debug overlays.

### Out of Scope

- Xbox Live integration.
- Online profile syncing.
- Global achievement account services.
- Achievement popup/toast notifications, unless added as a later follow-up
- Rich icon rendering unless added as a later follow-up

### Modified Dependencies

Existing Libraries only:
- ImGui, CLI11, and ReXGlue runtime
- XdbfWrapper
- XGI/XAM app handling
- CMake UI/runtime targets

### Acceptance Criteria

[...]

PR #282, opened 2026-04-12
feat(xam): Xbox 360 achievement system, Gamerprofile support

etonedemid:
## Xbox 360 Achievement System

PR #361, opened 2026-06-14
[Feature] Achievements Support

Graine25:
# Achievements 

This adds achievement support to the SDK. You pull a title's achievements (text + icons) out of its XDBF at build time, and at runtime they show up in an F7 overlay with unlock toasts. Unlocks are tracked live off the game's own XGI/XAM calls and saved per-title.

On top of the built-in behavior there's a small SDK API so recomp projects can define their *own* achievements, fire them from a hook, and swap out the icons or the entire ImGui UI when the defaults don't fit. That customization side is really the big apple of this PR. The built-in overlay is meant to be a sensible default, not the only option.

This is a new feature, nothing existing changes shape.

## How it fits together

- Authoring (CLI): `rexglue init achievements` pulls metadata + icons out of the title's XDBF into an output folder, normally the `metadata/` folder you ship next to the game data.
- Runtime (`rexsystem`): an `AchievementManager` holds the catalog, unlock state, persistence, and notification callbacks. `KernelState` loads it at module load; `Runtime` exposes it to the app.
- Guest hooks (`xam`): the XGI/XAM paths report unlocks into the manager and report current state back to the game.
- UI (`rexui`): the F7 overlay and the unlock toast, with PNG icons decoded through stb_image and cached.

Full write-up for the TOML format, path resolution, and API reference can be discussed internally.

## Authoring & replacing metadata

```
rexglue init --project-name <name> --xex-path <path/to/default.xex> achievements <path/to/default.xex> <path/to/output/>
```

Writes `achievements.toml` plus `icons/<image_id>.png` for every achievement and `icons/title.png`, the game's own icon dumped from the same XDBF. If you pass `<path/to/output/>` as `metadata/`, that becomes `metadata/achievements.toml` and `metadata/icons/...`.

Re-skinning an achievement is just editing files: change `label` / `description` / `gamerscore` in the TOML, or drop a different PNG in `icons/`. At load time, entries with the same `id` override the extracted ones, so your edits win over the defaults. Icons resolve in order: explicit `icon_path` => `icons/<image_id>.png` => the embedded XDBF image. Default discovery checks `<game_data_root>/metadata/`, then a sibling `metadata/`, then `<game_data_root>/`; an explicit `--metadata_root` overrides that search.

## Making it your own

This is the part aimed at recomp projects.

**Custom achievements**: Define one and unlock it from anywhere, including a guest hook. The convenience facade in `<rex/system/achievements.h>` is null-safe and thread-safe, so it's fine to call straight from a `REX_HOOK`:

```cpp
#include <rex/system/achievements.h>

// register once, e.g. in ReXApp::OnPostLoadXexImage()
rex::system::RegisterAchievement({.id = 0x10001, .label = "Speedrunner", .gamerscore = 25});

// unlock it whenever the trigger fires
rex::system::UnlockAchievement(0x10001);   // toast shows automatically
```

If you need more than that (listing, callbacks, manual persistence, etc.) the full `AchievementManager` is right there via `kernel_state()->achievements()`.

**Replacing the icons or the UI**: The overlay and toast are both override points on `ReXApp`:

- `CreateAchievementsOverlay()`: return your own `ui::ImGuiDialog` (or `nullptr` to turn the overlay off). The F7 keybind and toggle lifecycle are handled for you.
- `CreateAchievementNotificationDialog()`: swap the toast renderer for your own.
- Or build a completely custom window: pull data from `achievements()`, reuse `AchievementIconCache` for the icons, and draw whatever you want with `imgui_drawer()`: e.g. a "Trophies" page opened from a button inside your title. The built-in `AchievementsOverlayDialog` is there to copy from.

## What's in here

**CLI / authoring**
- `0f43ac9` add `rexglue init achievements` XEX=>TOML extractor (metadata, per-achievement icons, and the game's `title.png`)

**Runtime / system**
- `2ee37c6` shared `AchievementInfo` metadata struct
- `8ae58fd` `AchievementManager` (catalog, unlock timestamps, persistence, notifications)
- `0e2e0a2` wire the manager into `KernelState` (loads TOML, falls back to XDBF)
- `a976e15` expose the manager through `Runtime` + thread `metadata_root`
- `aef5ed6` track unlocks and report state back to the game via XGI/XAM
- `b1cd675` convenience hook API (`RegisterAchievement` / `UnlockAchievement` / `IsAchievementUnlocked`)
- `8735af2` cvar: defer unknown config flags until their flag registers (so late-registered achievement cvars still pick up TOML values)

**UI**
- `723b7c8` PNG decoding via vendored stb_image (This is where most of the changes come from, don't get scared by the large diff)
- `adff959` achievement icon cache
- `7ac1e86` the F7 achievements overlay
- `f98f3de` unlock toast notifications
- `7ea317f` wire overlay + toast into `ReXApp`

**Tests**
- `d5fc604` `AchievementManager` unit tests + a `ReXApp` consumer-compile target

## Tested

Ran this end-to-end against refof (Spider-Man: Friend Or Foe) as a real consumer check. Added a custom "67" achievement (67 gamerscore) registered through the new API and unlocked from a volume hook when all four volume sliders read 67. It builds and links against this branch, the custom achievement shows up in the overlay alongside the title's own, and it unlocks through the same toast path as everything else. Adapting refof to the new overlay came down to a one-line constructor change.

image

Heavy testing is required for us to send something like this out, so I would like for more contributors to give insight on something like this, if you have access to a game with unique behavior to achievements. Fair bit of work is referenced from #282. All feedback is welcome on a rather significant addition to the SDK. Solves #350


Graine25:
Would like to say a big thanks to @birabittoh for letting me know of this bug. Achievements were set in English only, so now we have a mode that plays achievements depending on the `user_language` cvar, as well as adding another cArg to the metadata dump feature. Command can be ran like so:

```
rexglue init achievements <path-to-default.xex> --language <id>
```

Running this command without the language flag will dump metadata in English by default. Behavior also applies to the in-game achievements overlay and toast. Open to even more feedback

birabittoh:
image
Thank you so much! Can confirm it's working out of the box for Kameo: RePowered!

mzzvxm:
image

Thanks bro! it will be on LARecomp

src/system/kernel_state.cpp
```diff
@@ -251,6 +253,66 @@ util::XdbfGameData KernelState::module_xdbf(object_ref<UserModule> exec_module)
...
+}
+
+AchievementListenerHandle KernelState::RegisterAchievementUnlockCallback(
+    AchievementUnlockCallback cb) {
+  return achievement_manager_.RegisterUnlockCallback(
+      [cb = std::move(cb)](const AchievementEvent& event) { cb(event.achievement); });
+}
+
+void KernelState::UnlockAchievement(uint32_t id) {
+  (void)achievement_manager_.UnlockAchievement(id, AchievementNotification::kSuppress);
```
tomcl7: So a real in-game unlock updates the overlay and the save file but shows no toast. Only the convenience API (`rex::system::UnlockAchievement`, which defaults `show_toast=true`) toasts. The screenshot is the custom "67" achievement, which takes the toasting path, so this gap wouldn't have shown up in your test. The description ("unlocks are tracked live off the game's own XGI/XAM calls" alongside "unlock toasts") reads as though the game's unlocks toast. If suppressing them is deliberate, for instance to avoid a burst the first time a title re-reports everything it already earned, that's reasonable, but the description should say so. If it isn't deliberate, the core function of this whole PR isn't firing for actual achievements.

src/system/achievement_manager.cpp
```diff
@@ -0,0 +1,332 @@
...
+                ec.message());
+    return;
+  }
+
+  std::string content = "# Achievement unlock state - managed by ReXGlue runtime\n\n";
+  for (const auto& [id, filetime] : unlocked) {
+    content += fmt::format("[unlocked.{}]\nfiletime = {}\n\n", id, filetime);
+  }
+
+  std::filesystem::path temporary_path = save_path;
```
tomcl7, line 298: Two guest threads unlocking two different achievements at nearly the same moment both open the same `<title>.toml.tmp`, interleave their writes, and both rename it into place. The result is a torn file, and on the next launch `LoadUnlockState` hits a parse error, logs a warning, and returns with an empty map, so every recorded unlock is gone. The same path also does the blocking disk write on the calling guest thread on every unlock. The convenience header advertises all of this as safe to call from any thread, so it's worth either serializing the save or making the temp name unique per write.

tomcl7 (commented):
This is a big and genuinely careful addition, and the write-up makes it easy to follow. Most of the description checks out against the code: extraction writes the TOML plus per-image icons, the runtime layers TOML over the XDBF defaults so an id collision replaces the extracted entry, unlock state is keyed per title id, the overlay reads a locked copy through `ListAchievements()`, and the notification callback only pushes onto the toast queue so the ImGui and texture work stays on the render thread. `DispatchCallbacks` snapshots under the lock and runs callbacks after releasing it, so there's no re-entrant deadlock. Good foundation.

A few things either don't match the description or look like real problems, I left inline reviews on.

The rest are smaller:
- The guest achievement enumerator now iterates `ks->loaded_achievements()`, which hands back the catalog vector by reference. That is the one read path that skips the mutex (`ListAchievements()` returns a copy under the lock, everything else locks). `RegisterAchievement` can `push_back` and reallocate that vector under the lock, so a registration from a hook thread while the game is enumerating is a reference-invalidation read. In practice registration usually happens once at startup so the window is small, but the API invites doing it "from anywhere," and the thread-safe claim doesn't hold for this accessor. Returning a copy here closes it.
- The toast notification callback captures a raw pointer to the toast dialog, and callbacks run outside the lock, so during shutdown `OnDestroy` can free the dialog while an unlock dispatched from a guest thread is mid-`Push`. Shutdown-only, but it's a use-after-free.
- The toast queue ages each entry from its arrival time and never re-stamps it when it reaches the front, and `OnDraw` uses that same timestamp for both the expiry check and the visible entry. So two unlocks within `kDisplaySeconds` of each other leave the second already expired by the time the first finishes, and it gets dropped without ever drawing. Only the first toast shows.
- `escape_toml` in the extractor only escapes `"` and `\`. XDBF strings come back as raw length-delimited bytes, so a label or description containing a newline or control byte produces an invalid TOML basic string and the whole `achievements.toml` fails to parse at load. Worth escaping the control characters too.
- The new "replays values for flags registered after loading" cvar test doesn't actually exercise the replay. The deferred values are only applied when `g_init_done` is set, which only happens inside `cvar::Init()`, and no test calls it (the test resets that flag to false first). So the late registration skips the replay, `value` stays `"F7"`, and `CHECK(value == "F10")` is false. The test needs to drive `Init()`, otherwise the path it claims to cover isn't being run.

None of this changes the overall direction, which is good. The two I'd want sorted first are the toast-on-game-unlock question, since it's the main user-facing behavior, and the save race, since it's the one data-loss path. And your own call for more testing is well placed: the `XGIUserWriteAchievements` buffer layout is reverse-engineered from one title's logs (the comment says as much, id at offset 4, stride 8), so a second title with a different layout is exactly what would surface that.

Very nice work.

src/system/kernel_state.cpp
Graine25: KernelState::UnlockAchievement() now calls `achievement_manager_.UnlockAchievement(id, AchievementNotification::kShow);` for the first thread so real XGI/XAM game unlocks now take the toast path too.

src/system/achievement_manager.cpp
Graine25: `AchievementManager::SaveUnlockState()` now takes a dedicated `save_mutex_` before snapshotting and writing the save file, so two guest threads cannot write the same `<title>.toml.tmp` concurrently anymore

tomcl7 (approved):
Nice work.

PR #365, opened 2026-06-18
fix(codegen): recover function tail after a conditional bcctr

kronflux:
## Summary

The function scanner classifies **every** branch-to-CTR (`bcctr`/`bcctrl`) as an
unconditional terminator. A *conditional* branch-to-CTR (`BO != 20`, e.g. `bnectr`,
`beqctr`) only transfers control to `CTR` when its condition is met and otherwise
**falls through** to the next instruction — just like a conditional `bclr`
(`bnelr`/`beqlr`), which the scanner already handles correctly.

Because the fall-through is never discovered, any function whose body contains a
conditional `bcctr` followed by more code is **truncated at the branch**, and the
entire fall-through body is silently dropped from the generated C++.

## Affected versions

Observed on SDK `0.8.0`. In `src/codegen/function_scanner.cpp`, two classification
sites in the block-discovery path (`discoverBlocks`, the function used by
`phase_discover`) are opcode-only and ignore the `BO` field:

- `isBlockTerminator()` returns `true` for **any** `bcctr`.
- `discoverBlocks()`'s explicit `bcctr` handler unconditionally `break`s the linear
  scan (before reaching the conditional-branch case that records a fall-through).

(`ppc::is_terminator_instruction(Opcode)` is opcode-level and cannot see `BO`; the two
`isTerminator()` call sites already guard with `&& !isConditional(...)`. The
conditional-`bclr`/conditional-`bc` paths in `discoverBlocks()` are the existing,
correct precedents this fix mirrors.)

## Root cause

`Opcode::bcctr` encodes both the unconditional `bctr` (`BO == 20`) and the
conditional forms. `Instruction::is_conditional()` already returns `XL.BO != 20`
for `bcctr`/`bcctrl`, but neither truncation site consults it.

## Reproduction

A common guest pattern — a hook/callback dispatcher that conditionally tail-calls a
registered handler and otherwise runs an inline body:

```asm
dispatch:
    lis     r4, g_hook@ha
    lwz     r0, g_hook@l(r4)   ; r0 = *g_hook  (a function pointer, may be null)
    cmpwi   r0, 0
    mtctr   r0
    bnectr                     ; if r0 != 0: tail-call the hook, return its result
    ; --- conditional fall-through (r0 == 0): a real subroutine body that the
    ;     recompiler DROPS. Begins mflr r0; mfcr r4; ... and ultimately returns. ---
```

Generated C++ (before this change) stops dead at the `bnectr`:

```cpp
if (!ctx.cr0.eq) { REX_CALL_INDIRECT_FUNC(ctx.ctr.u32); return; }
}   // <- function ends here; the entire fall-through body is missing,
    //    and r3 is left = the (non-zero) input argument
```

With the hook pointer null (its normal state), the truncated function returns its
non-zero input instead of taking the real fall-through path, corrupting downstream
control flow. The same executable runs correctly under a reference Xbox 360
interpreter, which executes the dropped body.

## Fix

In `function_scanner.cpp` (`discoverBlocks` path):

1. `isBlockTerminator()` — an unconditional `bcctr` (`bctr`, `BO == 20`) is a
   terminator; a conditional `bcctr` is not. Return
   `insn.opcode == Opcode::bcctr && !isConditional(insn)`.
2. `discoverBlocks()` — handle a **conditional** `bcctr` like a conditional branch:
   record the `addr + 4` fall-through label, push it onto the worklist, and continue
   the linear scan instead of `break`ing. Unconditional `bctr` (jump-table / tail
   dispatch) handling is unchanged.

After the fix, codegen recovers the full function (the dropped fall-through body):

```cpp
if (!ctx.cr0.eq) { REX_CALL_INDIRECT_FUNC(ctx.ctr.u32); return; }
ctx.r0.u64 = ctx.lr;     // mflr r0   (start of the previously-dropped body)
/* mfcr r4; ... rest of the real subroutine ... */
```

## Verification

- Built `rexglue` from this change and regenerated an affected module; the dispatcher
  function is no longer truncated — the full fall-through body is emitted (with no
  config size hint or hand-edit).
- The recompiled output boots and runs correctly with the corrected codegen.
- The change is symmetric with the already-correct conditional-`bclr` / conditional
  `bc` handling in the same function.

## Impact

Any function containing a conditional `bcctr` with reachable fall-through code is
affected (silent truncation → dropped logic / undefined return value). The fix is
narrow and does not change unconditional `bctr` / jump-table handling.


tomcl7 (approved):
Nice fix. Thank you!

PR #366, opened 2026-06-25
fix(ci): update Linux build dependencies

birabittoh:
Remove the now unused `wine` dependency and add `libxss-dev` to fix the build workflow.


tomcl7 (approved):
Thanks for the fix.

Issue #334, opened 2026-05-08
[Bug]: Vulkan: SPIR-V translator reads wrong fetch constant word for texture-fetch result exponent bias


ItsNotPaths:
### Description

in src/graphics/pipeline/shader/spirv_translator_fetch.cpp, the exponent bias for texture sampling is extracted from fetch_constant_word_4 (which is the LOD bias), not fetch_constant_word_3 where exp_adjust actually lives at bits 13:18. Every SPIR-V-translated texture fetch therefore scales its result by Ldexp(1.0, LOD_bias) instead of Ldexp(1.0, exp_adjust). the D3D12 translator is unaffected, its a vulkan-only bug. in fh1 it manifests as a fully-rendered scene with crushed mid-tones, geometry is correct but the entire image is dim, with only the very brightest HDR sources (sun, neon) surviving. The fix is the same patch xenia-canary landed in 32889f51b ("Use fetch_contant_word3 for exponent bias. fixes many graphic issues with Vulkan backend", 2025-12-06): load fetch_constant_word_3 alongside word_4 and use fetch_constant_word_3_signed in the BitFieldSExtract that feeds result_exponent_bias.

src/graphics/pipeline/shader/spirv_translator_fetch.cpp:
1) line 1411: add a load of fetch_constant_word_3 alongside the existing fetch_constant_word_4 load (same pattern)
2) line 2036: in the result_exponent_bias BitFieldSExtract call, swap fetch_constant_word_4_signed for fetch_constant_word_3_signed

### Steps to Reproduce

its provable by reading the translator and the X360 fetch-constant layout, no runtime repro is required, but the symptom is also visible at runtime in any title that sets a non-zero exp_adjust on a texture fetch.

### Expected Behavior

 Ldexp exp_adjust used instead of  Ldexp LOD_bias for texture fetches

### Actual Behavior

 Ldexp LOD_bias used instead of  Ldexp exp_adjust for texture fetches

### Platform

Linux

Graine25:
We're open to fixes on the Vulkan renderer. Edge and Canary both have updated fixes to their Vulkan backends, so if you plan on fixing your issues with fixes from those repositories, link the files that are changed in a PR 🙂

ItsNotPaths:
i wont pr because my fork has a bunch of other stuff and making said pr would be a pain but i can link to the xenia-canary commit - https://github.com/xenia-canary/xenia-canary/commit/32889f51beafc207ef3c4e73c807d363dfa133c3

PR #368, opened 2026-07-05
Fix Vulkan texture exponent bias, depth-only pipeline stalls, and WaitMultiple polling overhead

Ufoex:
## Summary

Three independent fixes found while debugging visual corruption and stuttering under the Vulkan backend, running Lollipop: Chainsaw via the Re-Cherry mod loader (Linux, NVIDIA proprietary driver):

1. **`spirv_translator_fetch.cpp`** - the SPIR-V texture-fetch translator read fetch constant word 4 for the result exponent bias instead of word 3 (D3D12's DXBC translator already reads word 3 for this). Caused blown-out brightness on affected texture fetches, Vulkan-only.

2. **`vulkan/pipeline_cache.cpp`** - async pipeline compilation required a pixel shader, so depth/shadow-only pipelines (no pixel shader) always compiled synchronously on the render thread. These are created continuously during gameplay (not just at load) whenever the camera reveals new shadow casters, causing render-thread stalls. Extended the existing "skip this draw, retry next frame" placeholder mechanism to cover this case (there's nothing cheaper to substitute for a depth-only pipeline, so it just queues the real one on the background thread pool and skips drawing with it until ready).

3. **`core/threading_posix.cpp`** - `PosixConditionBase::WaitMultiple` (backing `KeWaitForMultipleObjects`-style waits) polled every handle via `pthread_mutex_trylock` on a fixed 1ms interval regardless of wait duration. Any guest thread waiting on several kernel objects at once pays this cost continuously - measured via `perf`+Tracy on `AudioSystem`'s worker thread (waits on 9 handles), which dropped from ~68% to ~21% CPU after adding exponential backoff (1ms -> 16ms cap, resets per wait). Semantics/timeouts unchanged, only the poll interval.

## Test plan

- [x] Rebuilt RelWithDebInfo, confirmed no new warnings from the changed files
- [x] Ran Lollipop: Chainsaw via Re-Cherry on Linux/NVIDIA for multiple sessions after each fix
- [x] Fix 1: confirmed visually via screenshot that the exponent-bias corruption was gone
- [x] Fix 2: confirmed via log analysis that ~50% of newly created pipelines in a real session have no pixel shader; no crashes/hangs across repeated sessions after the change
- [x] Fix 3: confirmed via `perf record`/`perf report` before and after (WaitMultiple self-time 6.43% -> 2.73%, `pthread_mutex_trylock` 5.13% -> 2.30%) and `ps -T` thread CPU (Audio Worker 68% -> 21%); no hangs or audio glitches observed
- [ ] Not tested on D3D12/Windows or non-NVIDIA GPUs - would appreciate a sanity check there, especially for fix 3 since it's used well beyond the audio path

### DIFF
--- a/src/core/threading_posix.cpp
+++ b/src/core/threading_posix.cpp
@@ -281,6 +281,14 @@ class PosixConditionBase {
                         ? std::chrono::steady_clock::time_point::max()
                         : start_time + timeout;
 
+    // The poll interval backs off exponentially (capped) the longer nothing
+    // is signaled, so a long idle wait doesn't keep re-locking every handle's
+    // mutex ~1000 times/sec (measured as a real CPU cost on waits across
+    // several handles, e.g. AudioSystem's per-client semaphore WaitAny).
+    // Freshly started/likely-to-resolve-soon waits still poll at 1ms.
+    auto poll_interval = std::chrono::milliseconds(1);
+    constexpr auto kMaxPollInterval = std::chrono::milliseconds(16);
+
     while (true) {
       size_t first_signaled = std::numeric_limits<size_t>::max();
       bool condition_met = false;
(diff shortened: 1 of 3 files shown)


ItsNotPaths:
i can attest to 1, its a bad issue for fh1
https://github.com/rexglue/rexglue-sdk/issues/334#event-25456108667

Graine25:
Let's clean up the formatting errors

Graine25:
Feel free to re-PR once the merge issues are resolved

PR #370, opened 2026-07-08
fix(codegen): manual switch table configuration in block discovery

Graine25:
The manually configured switch tables in the config weren't being passed into `discoverBlocks()` at all. The call site in `phase_discover.cpp` simply didn't forward `ctx.Config().switchTables`.

The fix adds a manualSwitchTables parameter to `discoverBlocks()` so that user-configured tables (from `ctx.Config().switchTables`) take priority over auto-detection at each bcctr address. It also adds a guard -> `if (t != entryPoint && knownFunctions.contains(t)) continue;` <- to prevent jump table targets that land at the start of a known function from being incorrectly imported as basic blocks of the current function (i.e., tail dispatch is now handled correctly).

### DIFF
--- a/include/rex/codegen/function_scanner.h
+++ b/include/rex/codegen/function_scanner.h
@@ -253,10 +253,10 @@ struct BlockDiscoveryResult {
  * @param knownFunctions Set of known function entry points (to detect tail calls)
  * @return BlockDiscoveryResult containing blocks, branches, and jump tables
  */
-BlockDiscoveryResult discoverBlocks(DecodedBinary& decoded, uint32_t entryPoint,
-                                    const CodeRegion& containingRegion,
-                                    const std::unordered_set<uint32_t>& knownFunctions,
-                                    uint32_t pdataSize = 0);
+BlockDiscoveryResult discoverBlocks(
+    DecodedBinary& decoded, uint32_t entryPoint, const CodeRegion& containingRegion,
+    const std::unordered_set<uint32_t>& knownFunctions, uint32_t pdataSize = 0,
+    const std::unordered_map<uint32_t, JumpTable>* manualSwitchTables = nullptr);
 
 //=============================================================================
 // Jump Table Detection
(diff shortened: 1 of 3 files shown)


tomcl7 (changes requested):
The changes here are right. The discover phase never forwarded `ctx.Config().switchTables` to `discoverBlocks()`, so any manually configured table couldn't influence which blocks got walked. Preferring the manual entry at a given bctr address before falling back to `detectJumpTable`, keyed on the bctr address to match how config parses and stores it, is the correct place to hook it in.

One thing the description gets wrong, or at least doesn't make crystal clear is that manual switch tables weren't ignored before this. They were already consumed in the emit phase. the function graph looks them up twice, once to seed the label set:

https://github.com/rexglue/rexglue-sdk/blob/e8ce24fa73cd7c1ede80262c06f34893b7963dbe/src/codegen/function_graph.cpp#L405-L410

and again to pick the active table when emitting the bctr. So "weren't being passed into `discoverBlocks()` at all" is true only for the discover phase specifically. The actual failure is narrower and worth stating precisely: because discovery never saw the table, out-of-line case targets never became discovered blocks, so their `loc_XXXX:` labels were referenced by the emitted dispatch but never defined by any block. 

2 things on the change itself:

- The `knownFunctions` guard isn't scoped to manual tables. It sits inside the shared `if (jt)` block, so it now filters auto-detected table targets too. Previously every auto-detected target was imported as a block of the current function; now any target landing on a known function entry is dropped. That's a behavior change beyond what the title suggests. It's defensible (it makes jump table targets behave like the existing `isInternalTarget` check does for regular branches), but it should be called out as such, not folded silently into a manual-table fix.

- The comment calls manual tables "authoritative," then the guard immediately drops any user-listed target that happens to be a known function entry, with no log. If someone hand-writes a switch table in config and one of their targets got mis-registered as a function elsewhere, their explicit entry is discarded and they get no signal. For a user override that's described as authoritative, that silent drop is surprising. At minimum trace it so it's diagnosable.

Graine25:
I've decided to just add a couple comments on top of the warning/trace loggers with it

tomcl7 (approved):
Looks good. Nice job.

Issue #354, opened 2026-06-09
[Bug] FIFA World Cup 2010 - multi-XEX game crashes at unregistered guest function after DLL module loads


ripaldi2000-design:
### Description

1. Run `rexglue init` and `rexglue codegen --force` for FIFA World Cup 2010 South Africa (default.xex + fifadllzf.xex)
2. 2. Add fifadllzf.xex as a `[[modules]]` entry in the manifest with `guest_path = "fifadllzf.xex"`
3. 3. Build the project with `-DREXSDK_DIR` pointing to source tree
4. 4. Run: `fifa2010.exe --game_data_root=<path_to_game>`
5. 5. Observe crash in log

### Steps to Reproduce

a1. Run codegen on FIFA World Cup 2010 (default.xex + fifadllzf.xex)
2. Add fifadllzf.xex as [[modules]] in manifest
3. 3. Build and run fifa2010.exe --game_data_root=<path>
4. 4. Observe crash at 0x826581D0

### Expected Behavior

The DLL module loads and registers successfully, then the game continues execution past the XexLoadImage call.

### Actual Behavior

a[error] [cpu] Execute(82833420): function not in function table
[critical] [core] [FATAL] Call to invalid or unregistered function at guest address 0x826581D0

The address 0x826581D0 is within fifadllzf.xex's image range. The DLL is loaded and registered but functions from it are not being found in the dispatch table.

### Platform

Windows

### Clang version

22.1.7

### Version / Commit

v0.8.0

AKCore:
Not all functions will be found immediately, this is why you have the `[functions]` section in your toml files. Example i do something like this for forza4 

[[modules]]
guest_path = "XMediaFacade_default.xex"
file_path = "fm4/XMediaFacade_default.xex"
out_directory_path = "generated/xmedia"
includes = [
  "xmedia_config.toml",
]


and in the xmedia_config.toml


[functions]
0x88244AC0 = {} # pointer-referenced thunk; missed by discovery, called indirectly at runtime


discovery of Indirects are not the easiest thing to properly find especially with PowerPC assembly. 
You will go through a lot of indirects, this is just the nature of the beast. You will run, get a little further, get another [critical] [core] [FATAL] Call to invalid or unregistered function at guest address 0xXXXXXXXX  and then you will add that to your config. 

82xxxx is likely just part of your main config. so you can just do the [functions] declaration there. 

[...]

PR #371, opened 2026-07-08
fix(system): correct DLL code_base so indirect calls resolve correctly

Graine25:
solves #354 and other prototype codegen testing

### DIFF
--- a/include/rex/system/function_dispatcher.h
+++ b/include/rex/system/function_dispatcher.h
@@ -41,7 +41,8 @@ class ThreadState;
 class IModuleRegistrar {
  public:
   /**
-   * Returns false (and logs) if guest_address is outside every module range.
+   * Returns false (and logs) if guest_address is outside every registered
+   * function-table range.
    */
   virtual bool SetFunction(uint32_t guest_address, ::PPCFunc* func) = 0;
 
--- a/resources/templates/codegen/register_cpp.inja
+++ b/resources/templates/codegen/register_cpp.inja
@@ -19,3 +19,12 @@ void {{ project }}_RegisterFunctions(rex::runtime::IModuleRegistrar* registrar) 
 {% endif %}
 {% for fn in functions %}{% if not fn.below_code_base or fn.is_import %}  registrar->SetFunction({{ fn.address }}, {{ fn.name }});
 {% endif %}{% endfor %}}
+
+{% if is_dll %}
+// The runtime uses this baked layout for the dispatch table. It must match the
+// constants used by this module's REX_LOOKUP_FUNC macro.
+extern "C" REX_MODULE_EXPORT
+const rex::PPCImageInfo* ReXModule_GetImageInfo() {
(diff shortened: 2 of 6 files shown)


tomcl7 (changes requested):
Everything diagnosis wise seems to check out. Codegen computes the baked `REX_CODE_BASE` as the minimum start across all sections with `kXEPESectionMemoryExecute` set (via `BinaryView::fromModule`, which reads the flags `XexModule::PopulateBinaryData` copies out of the PE header), while the runtime side keyed the table on `.text` alone. The generated lookup indexes the table by the baked base:

```c
#define REX_LOOKUP_FUNC(x, y) \
  (*(PPCFunc**)(x + REX_IMAGE_BASE + REX_IMAGE_SIZE + (u64(u32(y) - REX_CODE_BASE) * 2)))
```

so when a DLL has executable sections before `.text`, the runtime writes slots at one offset and the DLL reads them at another, exactly as described. It's worse than just wrong resolutions, too: the runtime allocated the table sized from `.text` but the generated code indexes up to `REX_CODE_SIZE + REX_THUNK_RESERVE_SIZE`, so high targets read past the allocation entirely. The entrypoint path doesn't have this problem because `Runtime::Setup` consumes the baked `image_info.code_base` directly. This matches the #354 crash. The fix is correct.

A few things before I merge it:

The two new comments contradict each other. The `kernel_state.cpp` comment says DLLs place "executable import/export stub sections (.xidata/.xedata) *before* .text", which is the case the code_base fix handles. The `function_dispatcher.cpp` comment says import thunks in "non-executable sections (e.g. Xbox 'xidata')" sit below code_base, which is the only case where the new `guest_address >= mod->code_base` guard actually fires. Both can't be describing the same section in the same binary. Which one does the problem title actually hit? If its `.xidata` is marked executable, the dispatcher guard and the `FindModuleByAddress` widening never trigger for it and they're speculative; if it's not executable, the kernel_state comment's rationale is wrong even though the span change is still right for `.xedata` or whatever executable section sits low.

Smaller concerns:

- The runtime now re-derives code_base from PE sections, duplicating what codegen already computed and baked into the DLL. The two computations aren't identical: `BinaryView::fromModule` skips sections that are unmapped or extend past the image end, and your loop doesn't. A module with such an executable section would bake one base and compute another at runtime, which is this same bug again through a smaller door. The DLL already knows its own `REX_CODE_BASE`/`REX_CODE_SIZE`; is there a reason not to export them from the generated module (the way the entrypoint path already consumes the baked values) instead of recomputing?
- Widening `FindModuleByAddress` to start at `image_base` means `SetFunction` now accepts and reports success for any address in the image, including plain data addresses that get no fast-table slot. Previously those were rejected, which caught bogus registrations. Not a blocker given the memory side still bounds-checks itself, but it's a lost sanity check worth being aware of.

Once you cleanup the comments (and possible code additions) I will merge it.

tomcl7 (approved):

PR #375, opened 2026-07-11
Opt-in XAudio2 audio backend

Creesic:
A native XAudio2 audio backend as an alternative to the existing SDL-based backend. Handles guest→host format conversion (including an optional 5.1→stereo downmix with tunable per-channel coefficients), submits frames to XAudio2 on a dedicated MTA COM thread, and tracks playback rate against the guest's clock scalar via SetFrequencyRatio.

It's not a default addition, it's opt-in as its own target (rexaudio-xaudio2), excluded from the default build (EXCLUDE_FROM_ALL), and only compiled/linked when a downstream project explicitly asks for it:

rexglue_setup_target(my_game AUDIO_BACKENDS xaudio2)

That one argument does three things: links rexaudio-xaudio2 into the consumer's target, sets a REXGLUE_HAS_AUDIO_XAUDIO2 compile definition, and gates the xaudio2 branch and its #include in the shared audio_backend cvar dispatch. A project that doesn't opt in never sees the symbol, never compiles the backend, and its audio_backend cvar's allowed-values list doesn't even include "xaudio2". Default behavior (sdl) is unchanged either way.

### DIFF
--- a/cmake/rexglue_helpers.cmake
+++ b/cmake/rexglue_helpers.cmake
@@ -48,6 +48,8 @@ endfunction()
 #     so this single copy handles them transitively.
 #==========================================================
 function(rexglue_configure_target target_name)
+    cmake_parse_arguments(ARG "" "" "AUDIO_BACKENDS" ${ARGN})
+
     if(WIN32)
         target_sources(${target_name} PRIVATE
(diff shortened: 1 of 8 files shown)


Creesic:
Putting this on hold until https://github.com/rexglue/rexglue-sdk/pull/403 is resolved.

PR #379, opened 2026-07-19
vmsum qnan spec-correctness fix

ItsNotPaths:
fix(codegen): vmsum3fp128/vmsum4fp128 return QNaN on overflow, not inf

before this change we lower both to simde_mm_dp_ps, which computes in single precision and saturates to +/-inf when the dot product overflows. On Xenon these instructions return QNaN when normal doubles instead.

Repro: a vector whose components are +/-FLT_MAX.

    dir = { -FLT_MAX, -FLT_MAX, -FLT_MAX, -FLT_MAX }

    simde_mm_dp_ps(dir, dir, 0xEF)  ->  +inf (calculated in single, cannot tell if we got garbage infs in or calculated normally, 3fp/4fp do that)
    hardware                        ->  QNaN (0x7FC00000) (correctly emitting we didnt get garbage infs in but our result cant fit in a single float)

dp_ps accumulates in single precision, so it rounds at every add; 3fp/4fp rounds exactly twice, once in the accumulation and once in the final narrowing, improving accuracy (side effect). guest code using these instructions for collision or distance maths gets a more faithful result as well as the right overflow behaviour.

GJK termination tests commonly seed a FLT_MAX previous-distance term (before this change, to dp_ps) as an infinity sentinel and compare it against a newly computed squared distance. on VMX that squared distance comes from vmsum3fp128, and the seeding iteration overflows the single precision output. any guest code shaped that way depends on the overflow producing an unordered result when normal finite floats where input, and takes the wrong branch when we hand back an ordered +/-inf, in GJK's case terminating the solver immediately and discarding any collision it would have found.

3fp/4fp inputs are floats, so the largest possible product is FLT_MAX^2 ~ 1.2e77 and a four-term sum tops out near 4.6e77, against a double's 1.8e308 ceiling. the accumulator/3fp/4fp can't overflow when given normal input. an infinite accumulator therefore means an infinite input, and those pass through untouched. everything else narrows to float, and QNaN is substituted exactly when 3fp/4fp recieved finite floats and emitted a finite double that can't fit in a single. +/-inf is emitted only when one of the inputs was a inf, relaying garbage in > out. previouisly, dp_ps would emit +/-inf regardless of if it got inf in, or 2 normal single floats, since it only accumulated in single and couldnt tell.

addition is not associative and the order is observable under cancellation, so the reduction matches Xenia's: (x + z) + y for the 3-element form and (x + z) + (y + w) for the 4-element one. denormal results are flushed by hand, since the scalar path no longer inherits the host FTZ mode that the dp_ps lowering relied on.

Xenia implements this behavior on purpose. double accumulation plus an MXCSR overflow check that substitutes QNaN (x64_sequences.cc, DOT_PRODUCT_3_V128 and DOT_PRODUCT_4_V128), commented "this implementation is accurate, it matches the results of xb360 vmsum3", and enabled by default (use_fast_dot_product defaults to false). xenia also comments its own result is "often off by 1 bit" from hardware, so matching Xenia only. VMX128 semantics are not documented anywhere ive found. the free60 encoding list gives the bit layout and nothing about rounding, NaN or overflow, if someone has hardware or documentation to confirm or refute this, that is worth more than anything here.

The symptom that led here: Forza Horizon 1 has no collision between cars or props, which pass through each other, while terrain and world borders collide fine. the convex-v-convex narrowphase is the only path using GJK, and CGJK::Intersecting never returns true for any caller. its termination predicate computes

    cur  = vmsum3fp128(dir, dir)
    prev = previous distance squared, seeded to FLT_MAX
    terminate if (FLT_EPSILON * prev) >= (prev - cur)

and CSimplex seeds dir to -FLT_MAX so that the first dot product overflows by design. With QNaN the compare is unordered, all lanes are false, and the seeding iteration falls through so the solver runs. With +inf it becomes 4.06e31 >= -inf, which is true, so the predicate terminates on iteration 1 of every call and GJK never reports a hit.

verified against a Xenia build, at the predicate's blr it returns 0, both cr6-derived term results are zero, and the stored distance reads 0x7FC00000. this change reproduces those measured values (0x7FC00000 for the sentinel, 0x4009B8E9 for a real search direction) where dp_ps gave +inf, and restores collision.

### DIFF
--- a/include/rex/ppc/context.h
+++ b/include/rex/ppc/context.h
@@ -14,6 +14,7 @@
 #pragma once
 
 #include <bit>
+#include <cmath>
 #include <cstdint>
 #include <cstring>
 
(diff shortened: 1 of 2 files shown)


alligatorshoes:
> Note: vmsum4fp128 is changed for symmetry with vmsum3fp128 and with Xenia, which treats both identically. It is not on the path exercised above and was not measured, so that half is untested.

Tested on real hardware; it acts the same way, so you were right to change it for symmetry.

tomcl7 (changes requested):
This PR description is purely anecdotal with no technical merit. Please provide technical documentation and citations for what you are trying to change.

tomcl7:
I am not accepting this. Thanks for trying.

PR #382, opened 2026-07-21
fix(ppc): keep host FP exceptions masked in FPSCRRegister::setcsr

### LINKED ISSUE #381: [Bug]: aarch64: FPSCRRegister::setcsr can re-arm host FP exception traps (SIGFPE in guest code)
### Description

`rex::ppc::FPSCRRegister::setcsr()` writes the guest-derived control word straight to the host FP control register. `InitHost()` masks the host FP exception traps once at context init, but every later write goes through `setcsr()` unfiltered, so any csr value that arrives with the exception bits in their trapping state re-arms host FP traps for the remaining life of that context.

On aarch64 that is reachable in practice and the next guest FP op dies with SIGFPE. On x86_64 it is inert (see below), which is presumably why it has not shown up on linux-amd64.

The asymmetry is that the two `FPSCRPlatform` specialisations in `include/rex/platform/fpscr.h` use opposite polarity for the same field:

- x86_64 `ExceptionMask` bits are **mask** bits — `1 = masked`, and the MXCSR default already has all six set, so a zero-ish or partially rebuilt csr still lands masked.
- aarch64 `ExceptionMask` bits (FPCR IOE/DZE/OFE/UFE/IXE/IDE) are **enable** bits — `1 = trap enabled`. A stale value with any of them set arms the trap.

`InitHostExceptions()` already encodes this polarity per arch (`|=` on x86, `&= ~` on ARM); `setcsr()` just does not use it.

### Steps to Reproduce

1. Build for `linux-aarch64` (Clang 22, Release).
2. Run any title that executes `vcfsx` in recompiled guest code — the path that surfaced it here is `vcfsx` lowering to simde `_mm_cvtepi32_ps`.
3. Guest FP execution takes SIGFPE once a csr value with FPCR exception-enable bits set has passed through `setcsr()` (reached via `storeFromGuest()` / `enableFlushMode()` / `enableFlushModeUnconditional()`), rather than only via `InitHost()`.

Reproduced on Tegra X1 (Nintendo Switch / L4T) and Snapdragon SM8550 + Adreno 740 (AYANEO Pocket DS), both Linux aarch64.

### Expected Behavior

Host FP exception traps stay masked for the whole lifetime of a guest context. Guest code controls rounding mode and flush-to-zero through FPSCR; it must not be able to arm host traps, since the recompiler relies on FP status flags being sticky-and-silent rather than trapping.

### Actual Behavior

`SIGFPE` (FE_INVALID) from inside recompiled guest code on aarch64 after the first `setcsr()` that carries trapping-state exception bits.

Relevant code — `include/rex/ppc/context.h:195`:

```cpp
inline void setcsr(uint32_t csr) noexcept { Platform::setcsr(csr); }
```

vs. `InitHost()` at `include/rex/ppc/context.h:244`, which does apply `Platform::InitHostExceptions(csr)` first.

### Platform

Linux

### Clang version

22.1.8

### Version / Commit

v0.8.0 (`e8ce24f`)

- Fedora 44 aarch64
- Tegra X1 (Nintendo Switch, L4T) and Snapdragon SM8550 / Adreno 740 / Mesa Turnip (AYANEO Pocket DS)
- Also built and syntax-checked on x86_64 Fedora 44; the proposed fix is a no-op there.

### Related

- `include/rex/ppc/context.h:195` — `FPSCRRegister::setcsr`
- `include/rex/platform/fpscr.h:48,80` — `FPSCRPlatform::InitHostExceptions`, the two polarities
- I have a one-line fix for this (route `setcsr()` through `InitHostExceptions()`); happy to open a PR against this issue.

Azkali:
Fixes #381.

> **Corrected.** An earlier version of this description claimed this bug surfaces as SIGFPE on Tegra X1 and Adreno hosts. I measured it and that is wrong — see [the correction on #381](https://github.com/rexglue/rexglue-sdk/issues/381#issuecomment-5034112409). Neither of those platforms implements FPCR trapping. The change below still stands, but as hardening for hosts that do, not as a fix for an observed crash on those two.

## Summary

`FPSCRRegister::InitHost()` masks the host FP exception traps once at context init, but every later write to the control word goes through `setcsr()`, which passed its argument straight to `Platform::setcsr()`. A csr value reaching `setcsr()` with the exception bits in their trapping state therefore re-arms host FP traps for the rest of that context's life — reachable via `storeFromGuest()` / `enableFlushMode()` / `enableFlushModeUnconditional()`.

This routes `setcsr()` through `Platform::InitHostExceptions()`, the helper `InitHost()` already uses, instead of open-coding a mask that is correct on only one arch.

## The asymmetry

The two `FPSCRPlatform` specialisations in `include/rex/platform/fpscr.h` give the same field opposite meanings:

| | `ExceptionMask` bits | Default state |
|---|---|---|
| x86_64 | **mask** bits, `1 = masked` | MXCSR default has all six set |
| aarch64 | **enable** bits, `1 = trap enabled` | FPCR default has them clear |

`InitHostExceptions()` already handles this (`\|=` on x86, `&= ~` on ARM). `setcsr()` just did not use it.

## Scope — measured, not assumed

Trapped FP exceptions are **optional** in ARMv8-A; where unimplemented the FPCR enable bits are RAZ/WI. Setting `FPCR.IOE|DZE` and executing `0.0/0.0`:

| Host | CPU | Bits stick? | Result |
|---|---|---|---|
| Switch, L4T 4.9.140 | Tegra X1, Cortex-A57 | No — RAZ/WI | NaN, no signal |
| Pocket DS, Fedora 44 | SM8550, Cortex-X3/A715 | No — RAZ/WI | NaN, no signal |
| Apple M4, Fedora aarch64 | Apple M4 | Yes — `0x300` | **SIGFPE** |
| Apple M4, macOS | Apple M4 | Yes — `0x300` | **SIGILL** |

So this matters on hosts implementing the optional trapping behaviour (Apple Silicon today, where on macOS it presents as SIGILL rather than SIGFPE). On the ARM cores I can test it is a **latent hazard, not an observed bug** — nothing currently writes the enable bits, since `storeFromGuest()` only touches the rounding field. On x86_64 it is a no-op: the MXCSR mask bits are already set.

Cost: one ALU op per `setcsr()`.

## Test plan

- [x] Builds clean; header syntax-checks standalone with `clang++ -std=c++23 -fsyntax-only` on x86_64 and aarch64.
- [x] FPCR trapping behaviour characterised on four hosts (table above).
- [ ] Pocket DS (aarch64): confirm no regression in rounding-mode / flush-to-zero behaviour — guest still controls RMode/FZ, only the trap-enable bits are forced.
- [ ] linux-amd64: confirm no behavioural change.

If you consider hardening against a currently-latent issue out of scope, that's a fair call — happy for this to be closed.

## Notes

First of the changes from the closed #346, resubmitted one issue at a time as requested. The rest will follow as separate issues + PRs once this has been looked at.

### DIFF
--- a/include/rex/ppc/context.h
+++ b/include/rex/ppc/context.h
@@ -192,7 +192,18 @@ struct FPSCRRegister {
   static constexpr size_t FlushMask = Platform::FlushMask;
 
   inline uint32_t getcsr() noexcept { return Platform::getcsr(); }
-  inline void setcsr(uint32_t csr) noexcept { Platform::setcsr(csr); }
+  inline void setcsr(uint32_t csr) noexcept {
+    // Keep host FP exceptions masked on every write, not just in InitHost.
+    // Recompiled guest code reaches setcsr via storeFromGuest /
+    // enableFlushMode / etc., and a csr value that still has the exception
+    // bits in their trapping state (e.g. a context whose fpscr.csr field was
+    // zero-initialized before InitHost ran) would otherwise re-arm host FP
+    // traps and raise SIGFPE on the next FP op. InitHostExceptions applies the
+    // per-arch polarity (x86: 1 = masked, ARM: 1 = trap enabled), so it is used
+    // here rather than open-coding a mask that is correct on only one of them.
+    Platform::InitHostExceptions(csr);
+    Platform::setcsr(csr);
+  }
 
   inline uint32_t loadFromHost() noexcept {
     csr = getcsr();


mrcmunir:
Can you send this to the development branch? Thanks you .

mrcmunir (approved):

tomcl7:
@Azkali are you still finishing the remaining tasks for this or is it ready for final review?

PR #384, opened 2026-07-21
fix(ui): load config before reading the path cvars it can set

### LINKED ISSUE #383: [Bug]: config file is loaded after the path cvars it can set are read, so game_data_root in the TOML is ignored
### Description

`ReXApp::SetupEnvironment()` reads the four path cvars — `game_data_root`, `user_data_root`, `update_data_root`, `cache_path` — and bakes them into `PathConfig` *before* `rex::cvar::LoadConfig()` is called. A value set in the TOML config file therefore cannot influence path resolution: at the moment each cvar is read it still holds its default (empty).

The practical result is that `game_data_root` in the config file has no effect, and the app aborts with `--game_data_root was not provided.` — but only *after* it has initialised the presenter and created a swapchain, so the failure surfaces late and looks unrelated to config loading.

In `src/ui/rex_app.cpp` at v0.8.0 (`e8ce24f`), `SetupEnvironment()` begins at line 77:

```cpp
 81:  std::string game_data_cvar   = REXCVAR_GET(game_data_root);
 88:  std::string user_data_cvar   = REXCVAR_GET(user_data_root);
 97:  std::string update_data_cvar = REXCVAR_GET(update_data_root);
104:  std::string cache_path_cvar  = REXCVAR_GET(cache_path);
...
111:  PathConfig path_config{game_dir, user_dir, update_dir, cache_dir,
112:                         exe_dir / (std::string(GetName()) + ".toml")};
113:  OnConfigurePaths(path_config);
...
119:  resolved_defaults_ = std::move(path_config);
120:
121:  // Load config FIRST so log cvars have final values
122:  if (std::filesystem::exists(config_path_))
123:    rex::cvar::LoadConfig(config_path_);
```

The comment on line 121 is accurate for its stated purpose — the logging setup that follows at 126-139 does get final values — but the call sits below every path cvar read, so the "FIRST" only holds relative to logging, not to the paths above it.

There is no later recovery. `OnInitialize()` (line 64) passes the already-frozen `resolved_defaults_` to `OnFinalizePaths()`, whose default implementation (`include/rex/rex_app.h:128-132`) returns `defaults` unchanged, and `ConstructRuntime` then rejects it:

```cpp
166:  if (paths.game_data_root.empty()) {
167:    auto msg = std::string("--game_data_root was not provided.");
```

A subclass that overrides `OnConfigurePaths()` or `OnFinalizePaths()` and re-reads the cvars itself would paper over this, so it is only fatal on the default, unoverridden path — which is exactly the path that line 167 then rejects.

### Steps to Reproduce

1. Build any `ReXApp` consumer at v0.8.0 (`e8ce24f`).
2. Put the game path in the app's TOML next to the executable rather than on the command line:
   ```toml
   game_data_root = "/path/to/game"
   ```
3. Launch with no `--game_data_root` argument.

### Expected Behavior

`game_data_root` set in the config file is honoured, exactly as passing `--game_data_root` on the command line is. The config file and the command line should be interchangeable for these settings.

### Actual Behavior

The config value is ignored. The app initialises Vulkan, creates a swapchain, and only then aborts:

```
--game_data_root was not provided.
```

Passing `--game_data_root=...` on the command line works, because the CLI parse happens before `SetupEnvironment()` runs.

### Platform

Linux

### Clang version

22.1.8

### Version / Commit

v0.8.0 (`e8ce24f`)

Confirmed by reading the code at `e8ce24f`; the ordering is platform-independent and involves no arch-specific behaviour. Observed in practice on Fedora 44 aarch64.

### Related

- `src/ui/rex_app.cpp:81-123` — the reads and the load
- `include/rex/rex_app.h:128-132` — `OnFinalizePaths` default returns `defaults` unchanged
- `src/ui/rex_app.cpp:166-167` — where it aborts

I have a fix for this (resolve the default config path up front — it derives only from the executable folder and the app name, so the ordering cannot become circular — and load it before the cvar reads). Happy to open a PR against this issue.

Azkali:
Fixes #383.

## Summary

`ReXApp::SetupEnvironment()` read the four path cvars (`game_data_root`, `user_data_root`, `update_data_root`, `cache_path`) and baked them into `PathConfig` before `rex::cvar::LoadConfig()` ran, so a value set in the TOML could never influence path resolution — the cvar still held its empty default at the moment it was read.

This resolves the default config path up front and loads it before those reads.

## Why this is safe

The default config path is `exe_dir / (GetName() + ".toml")` — derived purely from the executable folder and the app name, with no dependency on any cvar. So loading it first cannot become circular, which is presumably why it was placed later in the first place.

`OnConfigurePaths()` may still redirect `config_path`, so the existing load is kept for exactly that case, now guarded on the path having actually changed to avoid applying the same file twice.

## Ordering, before and after

| | before | after |
|---|---|---|
| resolve default config path | line 111 (inside `PathConfig`) | up front, line 79 |
| **load config** | **line 123** | **line 83** |
| read `game_data_root` etc. | lines 81-104 | after the load |
| load redirected config | — | only if `OnConfigurePaths` changed it |

The comment that was on line 121 (`// Load config FIRST so log cvars have final values`) was accurate for the logging setup below it, but the call sat under every path cvar read — so "FIRST" only held relative to logging.

## Test plan

- [x] Builds clean at v0.8.0 + this patch.
- [x] Ordering verified by reading `e8ce24f`; the bug is platform-independent (no arch-specific behaviour involved).
- [ ] Confirm `game_data_root` set in the TOML is now honoured with no command-line argument.
- [ ] Confirm passing `--game_data_root` on the command line still wins where both are present, and that a config redirected by `OnConfigurePaths` still applies.

## Notes

Independent of the aarch64 work in #381/#382 — this one reproduces anywhere, no special hardware needed. Submitted as its own issue + PR per the one-at-a-time request.

### DIFF
--- a/src/ui/rex_app.cpp
+++ b/src/ui/rex_app.cpp
@@ -104,6 +104,10 @@ bool ReXApp::OnInitialize() {
 bool ReXApp::SetupEnvironment() {
   auto exe_dir = rex::filesystem::GetExecutableFolder();
 
+  auto default_config_path = exe_dir / (std::string(GetName()) + ".toml");
+  if (std::filesystem::exists(default_config_path))
+    rex::cvar::LoadConfig(default_config_path);
+
   std::filesystem::path game_dir;
   std::string game_data_cvar = REXCVAR_GET(game_data_root);
   if (!game_data_cvar.empty()) {
@@ -142,7 +146,7 @@ bool ReXApp::SetupEnvironment() {
   }
 
   PathConfig path_config{game_dir,  user_dir,     update_dir,
-                         cache_dir, metadata_dir, exe_dir / (std::string(GetName()) + ".toml")};
+                         cache_dir, metadata_dir, default_config_path};
   OnConfigurePaths(path_config);
   game_data_root_ = path_config.game_data_root;
   user_data_root_ = path_config.user_data_root;
@@ -152,8 +156,7 @@ bool ReXApp::SetupEnvironment() {
   config_path_ = path_config.config_path;
   resolved_defaults_ = std::move(path_config);
 
-  // Load config FIRST so log cvars have final values
-  if (std::filesystem::exists(config_path_))
+  if (config_path_ != default_config_path && std::filesystem::exists(config_path_))
     rex::cvar::LoadConfig(config_path_);
 
   // Late-phase logging


Graine25:
Formatting looks alright. Can you send this to the development branch?

Azkali:
> Formatting looks alright. Can you send this to the development branch?

Sure no problem !

PR #390, opened 2026-07-22
v0.9.0 - Achievements, SDL3 Windowing, and GPU Plugin Framework

tomcl7:
## Summary

This release adds a full achievement subsystem, replaces the platform specific window code with a single SDL3 backend, and moves Xenos GPU emulation out of the runtime into a plugin that loads at launch. It also carries a large safety pass over the platform and string layer, an interactive console with real commands, and the usual round of codegen, audio, and kernel fixes.

> [!CAUTION]
> This release has breaking changes. Migration steps are at the bottom.

### Achievements

- Add a runtime achievement subsystem that tracks unlocks and reports state back to the game.
    - `AchievementManager` lives on `Runtime` and is reachable through `KernelState`
    - The `xam` layer records unlocks and answers the game's queries for unlock state
    - Ships with an embedded metadata asset registry and a metadata store struct for per title data
- Add achievement overlays in the UI
    - Toast notifications on unlock, plus a full achievements overlay
    - Icon cache backed by PNG decoding through `stb_image`
    - Overlays are wired into `ReXApp`
- Add a convenience achievement hook API so titles and hooks can raise unlocks directly.
- Add `rexglue init achievements`, an extractor that reads achievement data from a XEX and writes it to TOML. 

Thank you @Graine25 for working on this!

### Windowing (SDL3)

- Replace the Win32 and GTK window backends with a single SDL3 backend. The old `window_win.cpp`, `windowed_app_context_win.cpp`, `windowed_app_context_gtk.cpp`, and the per platform `windowed_app_main_*` entry points are gone.
- Add window event hooks to `ReXApp` with a vetoable close, so an app can refuse or defer a window close request.
- Add a detached overlay mode for apps that drive their own renderer. The presenter can draw the overlay without owning the swapchain.
- Enable the UI dialog path on Linux builds.
- Add a per window cursor auto-hide delay.
- Default to fullscreen at launch.

### Graphics / GPU

- Extract Xenos GPU emulation into `rexgpu-xenos`, a plugin the runtime loads at launch instead of linking directly.
- Route kernel and app GPU access through the `IGraphicsSystem` interface so nothing downcasts to the Xenos implementation.

### Console Overlay

- Add an interactive console with command dispatch, name autocomplete, and a set of sample commands.
- Add argument passing commands and an `InvokeCommand` dispatch path so a command can take parameters, not just toggle a flag.

### Cvars

- Apply config file and command line values to cvars that register after runtime init, instead of dropping them.
- Defer unknown config flags until the flag that owns them registers, so ordering between a plugin's cvars and the config load no longer matters.

### CLI

- Print per binary title id, media id, version, and the PE file timestamp before each recompile.
- Gate the upgrade scan on the SDK version and skip generated trees so codegen output is not rescanned.
- Pin the reported version to the API floor value.

### Platform / Strings

- Add `rex::platform::env` and `rex::platform::console` wrappers and migrate call sites onto them.
    - Move `getenv` and `setenv` call sites to `rex::platform::env`
    - Route TTY detection and ANSI escape enabling through `rex::platform::console`
    - Move `kPathSeparator` into `rex::platform`
- Replace unsafe C string and formatting calls across the codebase.
    - Replace `strcpy`, `strncpy`, and `sprintf` with `rex::string` and `fmt::format_to_n`
    - Replace `gmtime_s` and `gmtime_r` with `fmt::gmtime`
    - Migrate `fopen` and `_wfopen` call sites to `rex::filesystem::OpenFile`
    - Consolidate the bounded copy primitives
- Collapse `system::SharedLibrary` onto `rex::platform::DynamicLibrary` so there is one dynamic library loader.

### Build / CMake

- Drop the direct `rexcore` link so it loads once rather than being pulled in twice.

### Logging

- Add `FlushLogging` to flush the loggers without shutting them down.
- Demote the VFS miss logs so they stop flooding the output.

## Fixes

- Correct the DLL `code_base` so indirect calls resolve to the right target. Thank you @Graine25!
- Configure the switch table manually during block discovery so codegen recovers the jump table. Thank you @Graine25!
- Recover the function tail after a conditional `bcctr` during codegen. Thank you @kronflux!
- Compute `vaddsws` with `blendv_ps` through casts so the saturating add is correct. Thank you @Graine25!
- Always overwrite generated output instead of skipping on a matching hash, so a stale file can never survive a regenerate.
- Byte-swap the message box title and body text so they render correctly
- Update the Linux build dependencies for CI/CD workflows
- Free the guest stack on `XThread` exit and terminate instead of holding it until object destruction.
- Decode kicked XMA contexts inline instead of blocking on the worker sweep.
- Fix the XMA loop path, which wrapped emitted garbage and dropped the loop-end frame.
- Flush the ffmpeg buffers on context release.
- Hide internal ffmpeg symbols for the Linux arm64 shared link.
- Ungate the `PPCContext` fields so its layout no longer depends on codegen optimization flags.
- Order the `xex2_version` bitfields for MSB packing.
- Cooperatively drain guest threads on title termination rather than force-killing them.
- Compare the raw owner word in the guest spinlock self-deadlock assert.
- Make `REX_IMPORT` callable static to prevent an ELF symbol collision with guest aliases.
- Restore monitor cvar window placement in the SDL3 backend.
- Write migration rewrites in binary mode so `CRLF` line endings are preserved.
- Miscellaneous console overlay fixes.

## Migration

There is no separate migration document, the breaking changes are covered here. The CLI's upgrade scan handles the renamed cvar and dropped config keys automatically. Run `rexglue codegen <project.toml>` directly to regenerate a project, and run the CLI itself rather than the cmake util target, because the upgrade detector is interactive and prompts for consent before applying changes, which a build target cannot handle.

Config and cvar renames the scan rewrites for you:

| Old name | New name | Kind |
| --- | --- | --- |
| `cache_path` | `cache_root` | cvar (renamed) |
| `xex_apply_patches` | (removed) | cvar |
| `patch_file_path` | (removed) | codegen config key |
| `patched_file_path` | (removed) | codegen config key |

Manual steps for host and app code:

- Deploy the `rexgpu-xenos` plugin next to the host executable. GPU emulation is no longer linked into the runtime, and the host loads the plugin at launch.
- Move any window code off the removed Win32 and GTK backends. There is one SDL3 backend now, so code that reached into `window_win.cpp` or the platform `windowed_app_*` files needs to go through the SDL3 path.
- Route GPU access through `IGraphicsSystem`. Code that downcast to the Xenos graphics system directly no longer compiles.
- Update any custom `Entry` subclass that overrides `RenameEntryInternal` to return `X_STATUS`, returning `X_STATUS_SUCCESS` on success and a failure code on a failed rename.
- Recompile projects so generated output picks up the `REX_IMPORT` static change and the ungated `PPCContext` layout.


Graine25 (approved):

Loreaxe (approved):
Looks all good here.

PR #391, opened 2026-07-23
feat(ui): add Wayland and VK_KHR_wayland_surface support

birabittoh:
Implements #317. I've been testing this on my [NocturneRecomp](https://github.com/birabittoh/NocturneRecomp) project for a few weeks and it's been working on many linux distros and desktop environments, including SteamOS.

* The feature commit contains the actual Wayland implementation
* The fix commit addresses an issue that many people ran into when playing my game on SteamOS: the virtual keyboard was popping up as soon as the game started, the special characters overlay was conflicting with letter inputs when trying to use keyboard controls. With this fix, these tools should only activate while the user is actually typing on an input field.

### DIFF
--- a/src/input/mnk/mnk_input_driver.cpp
+++ b/src/input/mnk/mnk_input_driver.cpp
@@ -270,6 +270,8 @@ void MnkInputDriver::CenterCursor() {
     ClientToScreen(hwnd, &pt);
     SetCursorPos(pt.x, pt.y);
   }
+#else
+  attached_window_->WarpMouseInWindow(cx, cy);
 #endif
 }
 
--- a/src/ui/imgui_drawer.cpp
+++ b/src/ui/imgui_drawer.cpp
@@ -371,6 +371,12 @@ void ImGuiDrawer::Draw(UIDrawContext& ui_draw_context) {
   SetupFontTexture();
 
   if (dialogs_.empty()) {
+    // No dialogs left to want text input (e.g. the last one was just
+    // removed); make sure the window doesn't stay flagged as a text field.
+    if (text_input_active_) {
+      text_input_active_ = false;
+      window_->SetTextInputActive(false);
+    }
     return;
   }
 
@@ -406,6 +412,12 @@ void ImGuiDrawer::Draw(UIDrawContext& ui_draw_context) {
   if (draw_data) {
     RenderDrawLists(draw_data, ui_draw_context);
   }
+
+  // Only mark the window as an active text input field while an ImGui
+  // text-entry widget is actually focused, so the OS/desktop input-assist
+  // UI (IME, autocomplete) doesn't kick in during normal button gameplay.
+  text_input_active_ = io.WantTextInput;
+  window_->SetTextInputActive(text_input_active_);
 
   if (reset_mouse_position_after_next_frame_) {
     reset_mouse_position_after_next_frame_ = false;
--- a/src/ui/surface_gnulinux.cpp
+++ b/src/ui/surface_gnulinux.cpp
@@ -28,5 +28,19 @@ bool XcbWindowSurface::GetSizeImpl(uint32_t& width_out, uint32_t& height_out) co
   return true;
 }
 
+#ifdef VK_USE_PLATFORM_WAYLAND_KHR
+#include <SDL3/SDL.h>
+
+bool WaylandSurface::GetSizeImpl(uint32_t& width_out, uint32_t& height_out) const {
+  int w = 0, h = 0;
+  if (!SDL_GetWindowSizeInPixels(sdl_window_, &w, &h) || w <= 0 || h <= 0) {
+    return false;
+  }
+  width_out = uint32_t(w);
+  height_out = uint32_t(h);
+  return true;
+}
+#endif
+
 }  // namespace ui
 }  // namespace rex
--- a/src/ui/window_sdl.cpp
+++ b/src/ui/window_sdl.cpp
@@ -221,6 +221,12 @@ void* WindowSDL::GetNativeWindowHandle() const {
 #endif
 }
 
+void WindowSDL::WarpMouseInWindow(int32_t x, int32_t y) {
+  if (sdl_window_) {
+    SDL_WarpMouseInWindow(sdl_window_, float(x), float(y));
+  }
+}
+
 uint32_t WindowSDL::GetLatestDpiImpl() const {
   float scale = sdl_window_ ? SDL_GetWindowDisplayScale(sdl_window_)
                             : SDL_GetDisplayContentScale(SDL_GetPrimaryDisplay());
@@ -246,10 +252,16 @@ void WindowSDL::ApplyNewTitle() {
 
 void WindowSDL::ApplyNewMouseCapture() {
   SDL_CaptureMouse(true);
+  if (sdl_window_) {
+    SDL_StopTextInput(sdl_window_);
+  }
 }
 
 void WindowSDL::ApplyNewMouseRelease() {
   SDL_CaptureMouse(false);
+  if (sdl_window_) {
+    SDL_StartTextInput(sdl_window_);
+  }
 }
 
 void WindowSDL::ApplyNewCursorVisibility(CursorVisibility old_cursor_visibility) {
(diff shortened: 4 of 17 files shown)


Graine25 (approved):

mrcmunir (approved):

src/ui/surface_gnulinux.cpp
```diff
@@ -28,5 +28,19 @@ bool XcbWindowSurface::GetSizeImpl(uint32_t& width_out, uint32_t& height_out) co
   return true;
 }
 
+#ifdef VK_USE_PLATFORM_WAYLAND_KHR
+#include <SDL3/SDL.h>
```
tomcl7, line 32: `SDL3/SDL.h` pulls in `string.h`, `wchar.h`, `math.h`, `stdint.h` and (under C++) libstdc++ headers. Including that at namespace scope declares everything it touches inside `rex::ui`, including any `namespace std { ... }` the libstdc++ wrappers open, which becomes `rex::ui::std`. It compiled for you because the include guards happened to already be satisfied by whatever came earlier in that translation unit. Change the SDL version or the include order and it stops. Move it to the top of the file with the other includes.

src/ui/window_sdl.cpp
```diff
@@ -161,8 +161,14 @@ bool WindowSDL::OpenImpl() {
...
-  // SDL3 requires explicit opt-in for text input events.
-  SDL_StartTextInput(sdl_window_);
+  // SDL3 requires explicit opt-in for text input events. We only want the
+  // character events (e.g. for ImGui overlays), not an OS on-screen keyboard
+  // popping up unprompted (observed on SteamOS/gamescope). Text input itself
+  // is left off until a text-entry widget is actually focused (see
+  // SetTextInputActive) so the window isn't flagged as an active text field
+  // for the whole session, which can trigger unwanted desktop input-assist
+  // UI (IME candidate windows, autocomplete) during normal gameplay.
+  SDL_SetHint(SDL_HINT_ENABLE_SCREEN_KEYBOARD, "0");
```
tomcl7, line 171: The comment says we only want the character events, then the code removes the only call that turns character events on, so the comment describes the opposite of what the line does. More to the point, the two halves of the fix are doing the same job twice. Once `SetTextInputActive` is driven by `WantTextInput`, the on-screen keyboard only appears when an ImGui text field actually has focus, which on a Steam Deck is exactly the moment the user needs it. Setting the hint to "0" on top of that means a Deck user with no physical keyboard can never type into `XamShowKeyboardUI` or the dev console at all. That's a capability loss, not a fix. Pick the gating, drop the hint.

Separately, `SDL_SetHint` is process-global but this is in `OpenImpl`, so it runs per window open. If you keep it, it belongs next to the other one-time SDL setup in `SDLWindowedAppContext::Initialize`.

src/ui/imgui_drawer.cpp
```diff
@@ -371,6 +371,12 @@ void ImGuiDrawer::Draw(UIDrawContext& ui_draw_context) {
   SetupFontTexture();
 
   if (dialogs_.empty()) {
+    // No dialogs left to want text input (e.g. the last one was just
+    // removed); make sure the window doesn't stay flagged as a text field.
+    if (text_input_active_) {
+      text_input_active_ = false;
+      window_->SetTextInputActive(false);
+    }
```
tomcl7, line 379: When the last dialog goes away, `DetachIfLastDialogRemoved` calls `presenter_->RemoveUIDrawerFromUIThread(this)`, so the presenter stops calling `Draw` entirely. The next `Draw` only happens when a new dialog is added, which may be never. So on the frame where the user closes a dialog that still had a focused `InputText`, the tail of `Draw` runs `text_input_active_ = io.WantTextInput` (true), calls `SetTextInputActive(true)`, then detaches, and the window stays flagged as a text field for the rest of the session. That's the exact state this fix exists to prevent. The clear needs to happen where the detach happens, not on a subsequent frame that won't come.

src/input/mnk/mnk_input_driver.cpp
```diff
@@ -270,6 +270,8 @@ void MnkInputDriver::CenterCursor() {
     ClientToScreen(hwnd, &pt);
     SetCursorPos(pt.x, pt.y);
   }
+#else
+  attached_window_->WarpMouseInWindow(cx, cy);
```
tomcl7, line 274: This is closing a gap (`GetNativeWindowHandle` returns nullptr on Linux, so `CenterCursor` was doing nothing there), but the only mouse grab in the codebase is here:
https://github.com/birabittoh/rexglue-sdk/blob/2912b35d0cffbb748d5806f5b03fde4afaa9a928/src/ui/window_sdl.cpp#L259-L261

`SDL_CaptureMouse` routes events outside the window, it doesn't lock or confine the pointer. Wayland compositors won't let a client move the pointer without a lock or confinement, so `SDL_WarpMouseInWindow` should be rejected on the native Wayland path. If that's right, mouse-look works on X11 and XWayland and drifts to the window edge on Wayland, which is the platform this PR is adding. Have you actually tested a mouse-look game under a native Wayland session, or has your testing gone through XWayland? If it needs relative mouse mode, that should be part of this PR rather than a follow-up, since Wayland is what's being enabled.

tomcl7 (changes requested):
A few things need to change before I can take it. I left in-line comments for you to review.

Smaller things:

- `target_compile_definitions(rexui PUBLIC VK_USE_PLATFORM_WAYLAND_KHR=1)` is PUBLIC but `target_include_directories(rexui PRIVATE ${WAYLAND_INCLUDE_DIRS})` is PRIVATE. Anything linking rexui gets the define, and `vulkan/api.h` then does `#include <wayland-client.h>` without the include path. It works today only because the header sits in `/usr/include` on most distros.
- The text-input commit adds `std::vector<uint8_t> icon_data_;` to `WindowSDL` with a comment saying it's applied in `OpenImpl`. `WindowSDL` doesn't override `LoadAndApplyIcon` and nothing reads the member. Drop it, or send it separately if there's an icon change coming.
- The commit message says `SDL_StartTextInput` was "re-armed on every mouse release". On `development` there is exactly one call to it, in `OpenImpl`. Nothing re-arms it. Worth correcting so the history matches the code.
- Opening the console now takes an extra frame before text input is live, since `SetTextInputActive(true)` runs at the tail of the first frame in which `WantTextInput` is set. Probably fine, but if a keystroke gets eaten on open that's why.

Please split this: the Wayland surface support and the text-input change are independent, and the text-input change touches every platform, not just Linux.

src/ui/surface_gnulinux.cpp
birabittoh: Moved to the top of the file with the other includes, guarded by `#ifdef VK_USE_PLATFORM_WAYLAND_KHR`, so nothing gets declared inside `rex::ui`.

src/ui/window_sdl.cpp
birabittoh: Agreed on both counts. The hint is gone entirely. `OpenImpl` no longer starts text input, `SetTextInputActive` drives it, and no `SDL_SetHint`. The comment in `OpenImpl` now explains why nothing is started there. Nothing process-global is left in `OpenImpl`, so there's no reason to move anything to `SDLWindowedAppContext::Initialize`.

src/ui/imgui_drawer.cpp
birabittoh: Moved the clear into `DetachIfLastDialogRemoved`, before the `RemoveUIDrawerFromUIThread` call, so it runs on the same path as the detach.

src/input/mnk/mnk_input_driver.cpp
birabittoh: I had tested this implementation on two different PCs: one of them running KDE Plasma and the other a minimal Wayland window manager.

The warp path **does** work. Mouse-look under a native Wayland session on KDE/KWin with the current `SDL_WarpMouseInWindow` code behaves correctly. Recent SDL3 appears to emulate the warp on Wayland rather than dropping it.

Anyway, it did feel hacky, so I've replaced it with `SetRelativeMouseMode` in my followup PR.

birabittoh:
Split done. I created #396 and #397 with the two fixed commits.

PR #396, opened 2026-07-31
feat(ui): add Wayland and VK_KHR_wayland_surface support with SDL3 backend

birabittoh:
Follow-up to #391, addresses #317.

Add a WaylandSurface surface type and the VK_KHR_wayland_surface instance extension, and stop forcing SDL_HINT_VIDEO_DRIVER=x11 when the Wayland path is compiled in, so a native Wayland session no longer has to go through XWayland.

Mouse-look re-centered the pointer every frame via SDL_WarpMouseInWindow, which on development was a no-op on Linux (GetNativeWindowHandle returns nullptr there). Warping does work under a native Wayland session with the SDL3 we link, which emulates it, but it depends on that emulation and on the compositor cooperating rather than on a mechanism the protocol actually provides. Replace it with Window::SetRelativeMouseMode, backed by SDL_SetWindowRelativeMouseMode, and carry the platform's relative motion through MouseEvent::dx/dy. This also removes the Win32 SetCursorPos special-case instead of adding a second platform branch beside it.

The mnk driver uses the relative deltas while the pointer is locked and falls back to absolute-position differencing if the platform refuses the lock, so it can't regress anywhere the old path worked.

### DIFF
--- a/.github/workflows/_build-platform.yaml
+++ b/.github/workflows/_build-platform.yaml
@@ -65,7 +65,7 @@ jobs:
           sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-noble.list http://packages.lunarg.com/vulkan/lunarg-vulkan-noble.list
           sudo apt update
           sudo apt install -y cmake ninja-build build-essential git curl unzip autoconf python3.12-venv \
-            libgtk-3-dev libx11-xcb-dev libxss-dev vulkan-sdk \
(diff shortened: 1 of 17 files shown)


tomcl7:
@birabittoh please resolve conflicts

birabittoh:
> @birabittoh please resolve conflicts

Done!

PR #397, opened 2026-07-31
fix(ui): only activate text input while an ImGui field is focused

birabittoh:
Follow up of #391.

SDL_StartTextInput was called unconditionally in WindowSDL::OpenImpl, permanently flagging the window as an active text field in SDL3. That makes the desktop's input assist kick in during normal button-only gameplay: an on-screen keyboard popping up at launch on SteamOS, and IME candidate windows / autocomplete on some desktop environments.

Add Window::SetTextInputActive and drive it from ImGuiIO::WantTextInput each frame, so SDL text input is only active while a text-entry widget (XamShowKeyboardUI dialog, dev console) is actually focused. This keeps the on-screen keyboard available where it's needed, which matters for Deck users with no physical keyboard.

Closing the last dialog detaches the drawer from the presenter, so no further frame runs; the flag is cleared in DetachIfLastDialogRemoved rather than on a subsequent frame that would never come.

### DIFF
--- a/include/rex/ui/imgui_drawer.h
+++ b/include/rex/ui/imgui_drawer.h
@@ -119,6 +119,11 @@ class ImGuiDrawer : public WindowInputListener, public UIDrawer {
 
   double frame_time_tick_frequency_;
   uint64_t last_frame_time_ticks_;
+
+  // Mirrors the last ImGuiIO::WantTextInput value applied to window_ via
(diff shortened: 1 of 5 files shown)


tomcl7:
@birabittoh please resolve conflicts

birabittoh:
> @birabittoh please resolve conflicts

Done!

PR #400, opened 2026-08-02
fix(gpu): Update outdated ZPD sentinel pairwise assumption

goldislead:
I'm aware that there's a preference to avoid any further Xenia renderer iteration, but there's an outdated assumption here about ZPD pending END detection that really should be updated.

Many, many titles never activate the ZPD B fields, so assuming that both ZFail & ZPass pairs must mirror each other is a bad idea.

Qualification: I wrote the Canary/Edge ZPD implementation.


Graine25:
Send to the development branch

goldislead:
> Send to the development branch

Done

PR #403, opened 2026-08-05
Audio stretch fix: keep SDL timer resolution enabled

Creesic:
The SDL audio setup was turning off Windows' high-resolution timer:

```cpp
SDL_SetHintWithPriority(SDL_HINT_TIMER_RESOLUTION, "0", SDL_HINT_OVERRIDE);
```
This is a one-character behavior change in `SDLAudioDriver::Initialize`. SDL documents `1` as its default on Windows. The problem came from the driver explicitly overriding that default with `0`.

I'm not sure how relevant the "_Prevent SDL from interfering with timer resolution (causes FPS drops)_" comment is here. I don't really encounter any problems. 

----------------------------------------

I had two builds, one with SDL and the other with XAudio2, for juxtaposition.

The test builds  logged:

- guest render callbacks and submissions;
- backend completions and queue depth;
- callback timing;
- source, stream, and device formats;
- FM2 scheduler, pump, and codec-read rates; and
- guest clock and output frequency ratios.

Both the broken and fixed builds submitted about 187 to 188 blocks per second.
That count alone made the SDL path look healthy.

The _contents_ were not healthy. Before the fix, more than half of the blocks were old data. Every duplicate matched the block from four submissions earlier.

-----------------------------------

| Run | Blocks checked | Unique | Repeated | Codec reads/s |
| --- | ---: | ---: | ---: | ---: |
| SDL before fix | 1,875 | 823 | 1,052 (56.1%) | 38-46 |
| XAudio2 | 1,875 | 1,875 | 0 | 91-94 |
| SDL after fix | 1,875 | 1,875 | 0 | 92-94 |

FM2 reads stereo PCM in 2,048-byte chunks:

```text
48,000 frames/s * 2 channels * 2 bytes = 192,000 bytes/s
192,000 / 2,048 = 93.75 reads/s
```

Before the fix, FM2 only read about 78 to 94 KB per second instead of the
required 192 KB. The missing data lined up with the percentage of repeated
audio blocks.

### DIFF
--- a/src/audio/sdl/sdl_audio_driver.cpp
+++ b/src/audio/sdl/sdl_audio_driver.cpp
@@ -37,10 +37,6 @@ SDLAudioDriver::~SDLAudioDriver() {
 }
 
 bool SDLAudioDriver::Initialize() {
-  // Guest workloads use millisecond-scale waits. Keep SDL's normal Windows
-  // timer resolution so those waits don't get rounded up.
-  SDL_SetHintWithPriority(SDL_HINT_TIMER_RESOLUTION, "1", SDL_HINT_OVERRIDE);
-
   // Set audio category for proper OS audio handling
   SDL_SetHint(SDL_HINT_AUDIO_CATEGORY, "playback");
 
--- a/src/audio/sdl/sdl_audio_driver.cpp
+++ b/src/audio/sdl/sdl_audio_driver.cpp
@@ -37,8 +37,9 @@ SDLAudioDriver::~SDLAudioDriver() {
 }
 
 bool SDLAudioDriver::Initialize() {
-  // Prevent SDL from interfering with timer resolution (causes FPS drops)
-  SDL_SetHintWithPriority(SDL_HINT_TIMER_RESOLUTION, "0", SDL_HINT_OVERRIDE);
+  // Guest workloads use millisecond-scale waits. Keep SDL's normal Windows
+  // timer resolution so those waits don't get rounded up.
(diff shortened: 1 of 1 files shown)


tomcl7:
I would prefer we just remove the hint instead of forcing it one way or the other. The comment you are referring to pre-dates ReXGlue and was likely brought in from Xenia.

src/audio/sdl/sdl_audio_driver.cpp
```diff
@@ -37,8 +37,9 @@ SDLAudioDriver::~SDLAudioDriver() {
 }
 
 bool SDLAudioDriver::Initialize() {
-  // Prevent SDL from interfering with timer resolution (causes FPS drops)
-  SDL_SetHintWithPriority(SDL_HINT_TIMER_RESOLUTION, "0", SDL_HINT_OVERRIDE);
+  // Guest workloads use millisecond-scale waits. Keep SDL's normal Windows
+  // timer resolution so those waits don't get rounded up.
+  SDL_SetHintWithPriority(SDL_HINT_TIMER_RESOLUTION, "1", SDL_HINT_OVERRIDE);
```
tomcl7:
```suggestion
```

Remove the explicit override instead of forcing one way or the other.

tomcl7 (changes requested):

Creesic:
Done and tested. Audio still good in FM2 with it entirely removed.

Graine25 (approved):

birabittoh:
I can confirm removing that single line completely fixes audio in The Darkness!

birabittoh (approved):

tomcl7 (approved):

PR #411, opened 2026-08-11
[Feature] Win-ARM64

Graine25:
This pull request adds support for ARM64 Devices on Windows. Instead of forcing users to run through a translation layer on Qualcomm devices, this PR will allow users to compile the SDK and then compile their project to run on the D3D12 Xenos backend (or replace the backend with a rendering system of choice) and execute code natively on the Microsoft ARM devices.

- **`5ac866d` win-arm64 ci**: Adds a `build-win-arm64.yaml` workflow and extends `_build-platform.yaml` with a six-way platform map so `win-arm64` runs on the `windows-11-arm` runner, alongside a nightly entry and a README update. The runner already ships native ARM64 LLVM, CMake, Ninja and VS 2022 with the MSVC ARM64 libraries, so the setup step only puts LLVM on `PATH` and prints versions for diagnostics rather than downloading a toolchain.

- **`4032013` adjust compiler targets for win-arm64 base**: Adds `CMAKE_C_COMPILER_TARGET` / `CMAKE_CXX_COMPILER_TARGET` of `aarch64-pc-windows-msvc` to `windows-arm64-base` in both `CMakePresets.json` and the `cmake_presets.inja` init template. Without the triple, clang would default to the host's `x86_64-pc-windows-msvc` and reject `-march=armv8-a`, so an ARM64-named preset would fail configuration or silently produce an x86 build.

- **`746970f` port windows exception handler register marshalling to arm64**: Wraps the existing x86 register marshalling in `REX_ARCH_AMD64` and adds a `REX_ARCH_ARM64` branch that copies `Pc`, `Sp`, `Cpsr`, `Fpsr`, `Fpcr` plus the `X` and `V` register files to and from `HostThreadContext`. Both `memcpy`s are guarded by `static_assert`s on matching sizes, so a future change to either the context struct or the Windows `CONTEXT` layout fails at compile time instead of silently corrupting register state.

- **`7e40ca6` include windows.h before pe_image.h in xex module**: Forces `<windows.h>` (with `WIN32_LEAN_AND_MEAN`) to be included ahead of `pe/pe_image.h` on all Windows targets. `pe_image.h` skips its own `IMAGE_*` definitions when it detects the SDK's via `#ifndef IMAGE_NT_SIGNATURE`; on win-arm64 nothing pulled in `<windows.h>` first, so it defined its own structs which then collided with `<winnt.h>` arriving later through the rex headers — win-amd64 happened to include it already and never hit this.

This is really a simplistic change and needs more work to wrap around those ARM-based systems. Significant amount of testing is required in order to approve this Pull Request. I have managed to get this to compile on the Github CI, as well as getting a template to compile, but I cannot confirm if this will work for a plethora of titles. Feedback is much obliged


furqanagwan:
Took the `xex_module.cpp` include-order change into a fork, because it is worth having on win-amd64 too and not only on arm64.

Which of `pe_image.h` and the Windows SDK defines the `IMAGE_*` structs is currently decided by whichever header reaches `<winnt.h>` first. On win-amd64 something always does, so it works - but by coincidence rather than by design, and the failure mode when the coincidence stops holding is a struct collision rather than anything that points at the cause. Making the include explicit costs nothing and removes the race. Builds clean here on x86-64.

One small thing: the local

```c
#ifndef WIN32_LEAN_AND_MEAN
#define WIN32_LEAN_AND_MEAN
#endif
```

is already covered - `src/system/CMakeLists.txt` has `target_compile_definitions(rexruntime PUBLIC NOMINMAX WIN32_LEAN_AND_MEAN)`. Worth knowing mainly because of what it does **not** define locally: `NOMINMAX`. This file uses `std::min`/`std::max` three times, so if that pattern gets copied into a translation unit outside `rexruntime`'s PUBLIC definitions it will pick up the `min`/`max` macros and fail in a way that reads like a template error. Might be worth either dropping the local define as redundant, or adding `NOMINMAX` beside it so the guard is self-contained.

I did not take the `exception_handler_win.cpp` half, only because it restructures a working AMD64 path for an architecture I cannot build or test here - no judgement on it intended.

PR #413, opened 2026-08-13
[Refactor] Vulkan-Xenos implementation to run with MoltenVK

Graine25:
refactor to the vulkan implementation we have in order to get it to run on macos via moltenvk. will edit readme momentarily


src/graphics/vulkan/pipeline_cache.cpp
```diff
@@ -3249,6 +3249,14 @@ bool VulkanPipelineCache::EnsurePipelineCreated(const PipelineCreationArguments&
   if (description.cull_back) {
     rasterization_state.cullMode |= VK_CULL_MODE_BACK_BIT;
   }
```
tomcl7, line 3251: Depth testing does not stand in for back-face culling. It does nothing when depth test or depth write is off, it does nothing for the transparent passes, and coplanar geometry that used to be culled now z-fights. On top of that, `EnsurePipelineCreated` fills `depth_stencil_state.front` and `depth_stencil_state.back` from separate guest registers, so every back face that used to be discarded now runs the back stencil ops and writes stencil. Anything doing stencil masking or shadow volumes gets different output, not just slower output.

tomcl7 (changes requested):
The rectangle-list SPIR-V work is the strongest part of this. Using the gl_PerVertex block type for a Function-storage variable really is invalid: the BuiltIn decorations sit on the type and are only allowed on Input and Output variables, so desktop drivers accepting it was luck. Splitting position and the distances into plain function-local arrays is the right call, the 8-entry bound matches `user_clip_plane_count : 3` plus `vertex_kill_and`, and the double-terminator guard in `SpirvBuilder::IfBuilder` is a real bug on its own, independent of MoltenVK.

Other concerns:

- 0x1002 is every AMD GPU, not just legacy ones, so this changes rendering on every AMD Mac including the 2019 Mac Pro and the 16-inch MacBook Pros. If MoltenVK is reporting a front face winding you don't expect, that reads like a winding problem to fix, not a reason to stop culling.
- `kLoadShaderIndexRGBA4ToRGBA8` gets a `load_shader_code` entry but never a `load_shader_code_scaled` one, while `kLoadShaderIndexR5G6B5ToRGBA8` gets both. `load_pipelines_scaled_` stays null for it, so with resolution scaling on, `IsScaledResolveSupportedForFormat` returns false and the load bails for k_4_4_4_4. Before the remap that format used `kLoadShaderIndexRGBA4ToARGB4`, which does have a scaled shader, so this is a regression on that path rather than a gap that was already there.
- The three new blobs in `vulkan_spirv/` ship as bare hex with no disassembly, while the other 93 files in that directory carry their full SPIR-V in the `#if 0` block. Nobody can regenerate these or read them to check, for instance, that the component order still matches the `RGBB` swizzle k_5_6_5 keeps after the host format swap to RGBA8.
- `-fexperimental-library` goes in via `add_compile_options` at the top of the root CMakeLists, so it applies to every Apple translation unit in the tree, MoltenVK and the loader included. The only thing that needs it is `std::jthread` in `src/core/timer_queue.cpp`. Put it on that target.

Graine25:
Culling tweak wasn't needed. Must've left it in there when I was trying out ways to fix things. Available to continue

tomcl7 (approved):

PR #417, opened 2026-08-16
refactor(rexglue): do not force `snake_case` project names

catuhana:
Previously, anything you passed to `--project_name` would be turned into `snake_case`, which may not be desired. Changing the `name` from manifest back to the desired name could also cause issues because of how `rexglue.cmake` was generated, e.g. re-creating the manifest file with `snake_case`, warnings (on Windows) while building.

I'm not sure if this is the desired approach, but it's the smallest change I could make. Should not be breaking since codegen phase re-generates `rexglue.cmake` and the manifest file.

Can't test on Linux right now, but should work.


furqanagwan:
Agreed on the motivation - forcing `snake_case` on a project name is unhelpful when the title is "Skate 3" or "Top Spin 4". One problem with routing `names.original` into the generated CMake, though: **`validate_app_name` deliberately allows spaces**, and CMake target names cannot contain them.

```cpp
if (!std::isalnum(c) && c != '_' && c != '-' && c != ' ') { ... }
//                                              ^^^^^^^^ spaces are valid input
```

So for an app named `Skate 3`, `cmakelists.inja` would generate:

```cmake
project(Skate 3 LANGUAGES CXX)
add_executable(Skate 3 WIN32 )
```

`project()` takes `Skate` as the name and then hits `3` where it expects a keyword, and `add_executable` treats `3` as a source file. Quoting would not rescue it either: `add_executable("Skate 3" ...)` is still an invalid target name. Names with a space are exactly the ones this PR is trying to support, so it would fail on its main case.

What seems to be conflated is the **display name** and the **identifier**. The display name is what belongs in `app_header.inja`'s comment and in the `ReXApp` constructor string, and `names.original` is right there. The identifier is what `project()`, `add_executable()` and the `_SOURCES` variable need, and that has to stay a sanitised form.

Keeping `snake_case` alongside the new `original` in `AppNameParts`, and using each where it belongs, would give you the readable names without the breakage. Worth a test with a two-word name before merging - the current templates would not configure.

(Separately: `parse_app_name`'s lambdas are marked `constexpr`, which for a lambda assigned to a variable means the closure is a constant expression, not that the calls are. Harmless, just not doing what it looks like it is doing.)

PR #425, opened 2026-09-01
feat(input): Implement Xbox 360 Input features

Graine25:
Brings the input layer up to parity with xenia-edge's `src/xenia/hid`, porting the device and configuration features it had that we lacked. The ported work is adapted to our device-centric architecture rather than copied — our `DeviceAssignment` and ordinal model stays, and Xenia's slot-binding table is deliberately not brought across. Verified end to end against reblue (RelWithDebInfo, driven under lldb), which surfaced three defects in the split that are fixed here.

### Edge-based refactor

- **`01caedf`**: Report the real `XINPUT_DEVSUBTYPE_*` instead of hardcoding `GAMEPAD`, so guitars, drum kits, wheels and dance pads are no longer locked out of the modes titles gate on subtype.
- **`c822729`**: Log every joystick arrival with VID/PID/GUID and warn when it has no gamepad mapping, so a controller SDL cannot promote no longer disappears silently (built to fix Steam Controller issues)
- **`1022130`**: Add a vibration master switch, radial stick deadzones, UI input blocking with held-button masking, and the `XN_SYS_INPUTDEVICESCHANGED` broadcast that titles poll for hotplug.
- **`3d74767`**: Add keyboard passthrough with USB HID usage codes and unicode, and make bound keys emit `VK_PAD_*` keystrokes, fixing a queue that was never filled.


Crafty-The-Fox:
This will fix hacks required in DJ Hero 2 (and other instruiment titles) to bypass controller type checks

PR #444, opened 2026-09-17
linux-amd64: fix libc++ primary template and host subdir SSSE3 flags

ArakiW:
# Upstream PR Description (ReXGlue SDK)

This file contains a ready-to-paste pull request description for the
`rexglue/rexglue-sdk` repository, including the two patches that fix
two independent build breakages on Linux x86_64 with Clang 20.1.8 and
libc++ 16+.

Both issues were discovered while porting a ReXGlue host build to a
Debian bookworm + Clang 20.1.8 + libc++ 16+ environment (the
`silkeh/clang:20` Docker image). They do **not** affect the upstream
Windows or Apple builds.

---

## Title

```
linux-amd64: fix libc++ primary template and host subdir SSSE3 flags
```

## Summary

Two minimal changes that together unblock a clean Linux x86_64 host
build with Clang 20.1.8 + libc++ 16+:

1. `include/rex/chrono/chrono.h` — extend the existing Apple-libc++
   workaround (`#ifdef __APPLE__`) to also cover Linux, FreeBSD, and
   Emscripten, whose libc++ does not expose
   `std::chrono::clock_time_conversion` as a primary template.
2. `src/core/CMakeLists.txt`, `src/audio/CMakeLists.txt`,
   `src/graphics/CMakeLists.txt` — call `rexglue_apply_target_settings()`
   on the host targets (`rexcore`, `rexaudio`, `rexgpu-xenos`) so they
   pick up the same `-msse4.1 -mssse3` flags that the host app and
   recompiled modules already get.

After these two changes, the host build completes with no extra
`-DCMAKE_CXX_FLAGS=-msse4.1 -mssse3` workaround.

## Test environment

- Docker image: `silkeh/clang:20` (= Debian bookworm + Clang 20.1.8 +
  libc++ 16+)
- Configure:
  `cmake -G Ninja -B out/build/cfg-clean -DCMAKE_BUILD_TYPE=Release -DREXGLUE_BUILD_TESTS=OFF -DREXGLUE_ENABLE_SANITIZERS=OFF -DREXGLUE_ENABLE_FIDELITYFX=OFF -DREXGLUE_ENABLE_TRACY=OFF -DREXGLUE_USE_VULKAN=ON`
- Build: `cmake --build out/build/cfg-clean --parallel $(nproc)`
- Result before this patch series: build fails at `src/core/memory.cpp`
  with `always_inline function '_mm_shuffle_epi8' requires target
  feature 'ssse3'`.
- Result after this patch series: build succeeds, **910/910 targets,
  0 warnings-as-errors, 0 errors**, `rexglue --help` runs successfully.

## Patch files (apply in this order)

```
docs/patches/0001-libcxx-clock-time-conversion.patch
docs/patches/0002-ssse3-host-subdir-helper.patch
```

Or with `git am`:

```
git clone https://github.com/rexglue/rexglue-sdk.git
cd rexglue-sdk
git checkout -b fh1-ssse3-libcxx-fixes
git am /path/to/0001-libcxx-clock-time-conversion.patch
git am /path/to/0002-ssse3-host-subdir-helper.patch
```

### Patch 1: `chrono.h` clock_time_conversion

Without the patch:

```
include/rex/chrono/chrono.h:131: error: explicit specialization of
    undeclared template struct 'clock_time_conversion'
include/rex/chrono/chrono.h:154: error: explicit specialization of
    non-template struct 'clock_time_conversion'
include/rex/chrono/chrono.h:154: error: redefinition of
    'clock_time_conversion'
```

The header already provides an empty primary template plus a `clock_cast`
helper under `#ifdef __APPLE__` for the same reason — Apple's libc++
does not expose `clock_time_conversion` as a primary template. Recent
Debian libc++ (16+) does the same, so the same workaround needs to
cover `__linux__`, `__EMSCRIPTEN__`, and `__FreeBSD__`.

### Patch 2: SSSE3 flags on host subdir targets

Without the patch:

```
src/core/memory.cpp:70: error: always_inline function '_mm_shuffle_epi8'
    requires target feature 'ssse3', but would be inlined into function
    'copy_and_swap_16_aligned' that is compiled without support for
    'ssse3'
src/audio/xma_context.cpp:775: error: ... (xma_context frame interleaving)
include/rex/audio/conversion.h:42: error: ... (pshufb-style byte swap)
include/rex/graphics/primitive_processor.h:424: error: ... (ReplaceResetIndex32To24)
```

`rexglue_apply_target_settings()` in `cmake/rexglue_helpers.cmake`
already sets `-msse4.1 -mssse3` on AMD64 hosts, but it is only invoked
from `rexglue_configure_target` (host app) and
`rexglue_configure_module_target` (recompiled guest modules). The host
subdirs that build their own targets from the SDK source tree never
call it, so the SSSE3 flag is never passed to those targets.

Calling `rexglue_apply_target_settings()` on each affected target
(`rexcore`, `rexaudio`, `rexgpu-xenos`) right after their setup is
complete picks up the same flags without requiring a global
`CMAKE_CXX_FLAGS` override at configure time.

## Out-of-scope (future work)

- `rexglue_apply_target_settings` should arguably be invoked by a
  higher-level `apply_host_target_settings()` helper that every host
  subdir calls by convention. That is a refactor; this PR only fixes
  the immediate build breakages.
- The same `rexglue_apply_target_settings` helper currently only adds
  SSE flags on x86_64; ARM64 builds would need an equivalent path
  (currently ReXGlue falls back to `REX_ARCH_ARM64` paths in the
  affected source files).

## Local verification

The FH1-Switch project on disk verifies this patch series end-to-end:

```
E:\AI\FH1_SWITCH\
├── .local\rexglue-sdk\            # clean main, then patch applied locally
├── docs\patches\0001-libcxx-clock-time-conversion.patch
├── docs\patches\0002-ssse3-host-subdir-helper.patch
├── docs\upstream\PR_DESCRIPTION.md # this file
├── tools\docker\build-clean.sh    # scripted reproduction of the test
└── README.md                       # workspace entry point
```

To reproduce:

```bash
docker build -f tools/docker/Dockerfile.fh1-rexglue-host -t fh1-rexglue-host:latest .
docker run --rm -v 'E:\AI\FH1_SWITCH:/work' fh1-rexglue-host:latest \
    bash /work/tools/docker/build-clean.sh
docker run --rm -v 'E:\AI\FH1_SWITCH:/work' fh1-rexglue-host:latest \
    /work/.local/rexglue-sdk/out/linux-amd64/rexglue --help
```


furqanagwan:
Took the CMake half of this into a fork and it is a real gap - `rexcore`, `rexaudio` and `rexgpu-xenos` never had `rexglue_apply_target_settings` applied, so they compile without `-msse4.1` while inlining `_mm_shuffle_epi8` in `copy_and_swap_*`, `xma_context.cpp` and `ReplaceResetIndex32To24`. Inert on MSVC, which is presumably why it went unnoticed on Windows for so long. Thank you for finding it.

One concern about the `chrono.h` half, which I did not take as written.

The guard becomes `defined(__APPLE__) || defined(__linux__) || defined(__EMSCRIPTEN__) || defined(__FreeBSD__)`, but whether `std::chrono::clock_time_conversion` exists is a property of the **standard library**, not of the platform. libstdc++ does provide the primary template (C++20 `clock_cast` has been there since GCC 11), and it is the default for GCC and for most distributions' clang. On that combination this patch declares

```cpp
template <class, class> struct clock_time_conversion {};
```

a second time in `namespace std::chrono`, which should be a redefinition error rather than a fix - so the change may trade a libc++ build failure for a libstdc++ one, on the same platform it is trying to fix.

`_LIBCPP_VERSION` seems like the condition you actually want:

```cpp
#if defined(_LIBCPP_VERSION)
```

It covers Apple and Linux libc++ together, drops the need to enumerate FreeBSD and emscripten, and leaves libstdc++ alone. `<chrono>` is already included above that point, so the macro is defined by then.

I could not test this - no Linux machine here, so I am reasoning from the libstdc++ side rather than reporting a failure I have seen. Worth a check against GCC or clang+libstdc++ before merging either way.

